PR #4, opened 2020-01-29
Selection Filters + small accessibility fixes

JBuma:
I mostly moved around a lot of stuff and changed some data so filtering on groups will be easier later on. Local and working groups are now in their own store (although they will be moved to the server, once we have one), and now follow a {label, key} format for easier searching and filtering.
Also changed some properties on the inputs for better accessibility (the old inputs removed labels when they had a value).


SamLubbers:
This pull request resolved #3

PR #10, opened 2020-02-09
Prettier setup

JBuma:
I installed prettier and eslint as dependencies with some basic setup to make them work nicely together. This way eslint won't give warnings about rules that conflict with prettier.
Check if your editor has a plugin for prettier and eslint. I use vscode, which has plugins for both [eslint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) and [prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode).


JBuma:
I don't particularly mind either way, so i moved the configs to `package.json`. Also modified the `"vue/max-attributes-per-line"` rule a little so it plays a little nicer with prettier formatting.

Issue #9, opened 2020-02-03
Setup basic user-side navigation


JBuma:
We will be supporting both tasks and roles in the first release, and we want them to be in separate lists to keep them distinct. Basic structure will be something like:

**Homepage:** {base_url}
**Task overview:** {base_url}/tasks
**Task detail:** {base_url}/tasks/{task_id}
**Task create:** {base_url}/tasks/new

Same thing for roles. Since this will be a Single Page Application (or SPA), navigation will take place on the frontend using [VueRouter](https://router.vuejs.org/). This is already installed and set up so you can get going immediately.

abudaan:
Okay, and then only {base_url}/tasks/new requires authentication / authorization?

SamLubbers:
@abudaan that is correct.

SamLubbers:
A few additional questions to answer about the URL structure:

In the (long term) future we aspire other national XR branches to also use this platform to manage their roles & tasks. To easily support this we should have the country code as the first element in the URL structure `{base_url}/{country_code}`. For example, for this first version only used in The Netherlands, we will have `{base_url}/NL`, `{base_url}/NL/tasks `...

In this version, the home page will be the role overview page. However, in the future, we might want a different home page. Either a global page where you can navigate to the pages of different countries, or a home page where users can see their roles / tasks. To support this we should keep clear the base URL: when users navigate to `/` they will be immediately redirected to `/roles`

SamLubbers:
### international URL
In the previous comment, where I talked about other countries also making use of this application, I had assumed that they would all host their roles in the same website, for example, on `xrvolunteer.com`. However, having all countries make use of a single domain has several problems:
  * security: if this single domain is brought down, the applications stops working for all countries
  * scalability issues: if the application is used worldwide, we must put in extra work for the application to scale properly, so it works fast wherever it is used.
  * energy use: transmitting data over larger distances requires more energy.

An improved alternative solution, is for each country to self-host their own version of the application. So the XR NL platform could be hosted on `volunteer.extinctionrebellion.nl` and the XR US platform could be hosted on `volunteer.extinctionrebellion.nl`. 

With this configuration, we don't need the country code in the URL, like I mentioned in the previous post.

### home page redirect
In the previous post I also mentioned redirecting from `/` to `/roles`. However, we will integrate roles and tasks into the same page, so we could instead redirect to `/explore`, where both roles and tasks will be displayed.

SamLubbers:
[...]

PR #11, opened 2020-02-09
Role detail view

### LINKED ISSUE #2: View details of individual role.
When a user clicks on a role card they should able to view all the details about the role.

The role details can be found in the wiki

JBuma:
Added some basic navigation for viewing the roles. It uses some nested routing so you still have a dialog, but can also directly link to it when needed. Since there were some issues with the routing i laid out in #9, i used roles/view/:id for viewing a single role. We'll probably have to sit down and think of how we want to actually do routing to avoid conflicts since we also may want to do filtering in the url as well sometime in the future.
The dialog component looks at the :id parameter in the url, and uses that to look up the actual role. Right now from the store, later from the server.
resolves #2

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -6,13 +6,13 @@
       https://router.vuejs.org/guide/essentials/dynamic-matching.html#reacting-to-params-changes
       :key ensures component is always re-rendered from scratch,
       eliminating the need to watch routes or navigation guards in components -->
-      <router-view :key="$route.fullPath" />
+      <router-view />
     </v-content>
   </v-app>
 </template>
 
 <script>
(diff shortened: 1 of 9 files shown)


src/App.vue
```diff
@@ -6,24 +6,23 @@
       https://router.vuejs.org/guide/essentials/dynamic-matching.html#reacting-to-params-changes
       :key ensures component is always re-rendered from scratch,
       eliminating the need to watch routes or navigation guards in components -->
```
SamLubbers: comment can be deleted now we do not use `:key` in this router-view

PR #12, opened 2020-02-09
GraphQL

abudaan:
The frontend is connected to Hasura server running on Heroku. See these live urls:

- [frontend](https://xr-volunteer-app.abumarkub.org/)
- [backend](https://xr-volunteer-app.herokuapp.com/console/data/schema/public)

The state is kept in Vuex. The queried data from the server is stored in the local GraphQL cache. So there is no single source of truth. I am working in another [branch](https://github.com/abudaan/aginun/tree/local-graphql) on a version that uses local GraphQL cache for the entire app state.


abudaan:
Closing this PR for now: I want to make a few changes first.

SamLubbers:
I added the [WIP](https://github.com/marketplace/wip) application so that we can we do not have to close open pull requests with work in progress.

abudaan:
Please don't waste any time on merging this PR: the new branch that I am working on has a complete different approach towards state management and GraphQL integration.

abudaan:
> same text with colored syntax highlighting on [hackmd](https://hackmd.io/vD_jKpLuQGqXgkzCDSd3RA) 

# Remarks

## Folders and files
1. All Apollo related code is in the `src/apollo` folder.
1. GraphQL queries that are used in multiple files are stored in separate `.gql` files.
2. The local GraphQL cache can be found in the file `src/apollo/vue-apollo.js`. In this file you can also find the settings of Apollo client
3. The local resolvers are all stored in their own file in the folder `resolvers`. All resolvers are made available to Apollo client via the file `src/apollo/resolvers.js`.
4. If you look at the resolvers object in this file: the resolvers of the fields `Query` and `Mutation` are accessible via the root, the resolvers in the field `RoleData` are only accessible via `getRoleData` under the `Query` field, see explanation below.

## General
1. For syntax highlighting of GraphQL files you can install this vs-code [plugin](https://marketplace.visualstudio.com/items?itemName=kumar-harsh.graphql-for-vscode). Note however that the auto-complete function is very buggy.
1. The build script chokes on `.gql` and `.graphql` files but the develop script (`npm run serve`) does not. This is probably solvable by a setting in the `vue.config.js`

## GraphQL
1. Wherever possible I have used server queries; this is used when the queried data doesn't need much or no post-processing.
2. When data does need post-processing I have used 3 solutions:
    1. Post-processing of the data in the `.vue` file
    2. Post-processing in a local resolver
    3. A combination of both
3. About post-processing in a local resolver: for this we need to overrule the default resolver that looks for fields in the local cache, see examples below.

## Some examples

Example of a query that retrieves the filter data directly from the cache data object.
```graphql=
 query GetFilter {
  roleData @client {
    filter {
      ...RoleFilterFragment 
    }
}
```

If you look at the structure of the cache data object you see why that is:
```js=
  data: {
    roleData: {
      id: "data",
      __typename: "RoleData",
      filter: {
        id: "filter",
        __typename: "RoleFilter",
```

However, if you want to perform some post- or pre-processing you have to bypass the default resolver and construct a query that uses `getRoleData`. If you take a look a the file `src/apollo/resolvers/getRoleData.js` you see that  it basically queries the complete `roleData` object from the cache. And if you query for a non-scalar field in your query, the resolver will move on to a resolver with the same name.

So consider a query like this:
```graphql=
 query GetFilter {
  getRoleData @client {
    filter {
      ...RoleFilterFragment 
    }
}
```
It will first run the `getRoleData` resolver and then the `filter` resolver in the file `src/apollo/filter.js`. Note that a subsequent resolver will *only* run if the queried field is non-scalar! If you need processing on a scalar field, make it non-scalar first, for instance:

```graphql=
type roleData {
    amount: Int!
}
```
becomes:
```graphql=
type roleData {
    amount {
        __typename: "Amount"
        value: Int!
    }
}
```

SamLubbers:
Instead of merging the +10.000 lines of code in this pull request, we will close this pull request and merge parts of this code in separate smaller pull requests. This will be easier to merge, lead to fewer errors, and make backend code available sooner in the master branch.

Because of the difficulties with using only `vue-apollo` with client side queries and a local cache, we will use a vuex store as a local cache and only make use of server side queries, eliminating the need for resolvers.

PR #30, opened 2020-02-19
Role creation

### LINKED ISSUE #1: Role creation
Implement all steps for the use case [#2: creating a role](https://github.com/xrnl/aginun/wiki/Functional-Specification#2-creating-a-role), except for the steps that require user authentication.

The design of the create dialog should be consistent with the dialog to view a role. This will make it easy to edit the role.

SamLubbers:
Form with validation to create a new role. Closes #1 

This pull request is still missing:
- responsibilities validation: minimum of 1 item and a maximum of 10
- showing which fields are required with a *
- tooltip on disabled submit button explaining that all roles must be filled.


src/components/NewItemDialog.vue
```diff
@@ -0,0 +1,289 @@
...
+  message: "The {_field_} must be under {length} characters.",
+});
+extend("phoneDutch", {
+  validate: value => {
+    const dutchPhoneRegex = RegExp(
+      /^((\+|00(\s|\s?-\s?)?)31(\s|\s?-\s?)?(\(0\)[-\s]?)?|0)[1-9]((\s|\s?-\s?)?[0-9])((\s|\s?-\s?)?[0-9])((\s|\s?-\s?)?[0-9])\s?[0-9]\s?[0-9]\s?[0-9]\s?[0-9]\s?[0-9]$/
+    );
+    return dutchPhoneRegex.test(value);
+  },
+  message: "You must enter a valid Dutch phone number",
```
JBuma: We might need to expand this to other phone numbers as well, not everyone has a dutch one.

src/components/NewItemDialog.vue
```diff
@@ -0,0 +1,289 @@
...
+                  :disabled="isEmpty(newResponsibility)"
+                  @click="addResponsibility"
+                >
+                  <v-icon>mdi-plus</v-icon>
+                </v-btn>
+              </template>
+            </v-text-field>
+          </validation-provider>
+          <v-card v-if="responsibilities.length > 0" class="mb-4">
+            <template v-for="(r, i) in responsibilities">
```
JBuma: Just personal preference, but i like writing out variables so it's clearer what they actually are. So `r` would just be `responsibility`. A bit longer to write out, but your IDE should have autocomplete anyway.

src/components/NewItemDialog.vue
```diff
@@ -0,0 +1,289 @@
...
+                  @click="responsibilities.splice(i, 1)"
+                >
+                  <v-icon>mdi-close</v-icon>
+                </v-btn>
+              </v-list-item>
+            </template>
+          </v-card>
+          <validation-provider
+            v-slot="{ errors }"
+            rules="required|max:1000"
```
JBuma: I don't think this should be required.

src/components/NewItemDialog.vue
```diff
@@ -0,0 +1,289 @@
...
+            </v-select>
+          </validation-provider>
+          <p class="caption mb-0" style="color: gray">
+            Contact details
+          </p>
+          <validation-provider
+            v-slot="{ errors }"
+            name="email address"
+            mode="eager"
+            rules="required|email"
```
JBuma: Maybe instead of always requiring an email adress, make it required that at least one of the contact details fields is filled, so you can choose between email, phone, etc.
Related to this, it might also be nice to have a general text field for contact details. Maybe there's a certain signal group people need to join, some sort of secret meeting location, maybe people prefer to use a platform we don't know about.

src/components/NewItemButton.vue
```diff
@@ -0,0 +1,14 @@
+<template>
+  <v-btn color="primary" @click.stop="$emit('click')">
+    <v-icon>
+      mdi-plus
+    </v-icon>
+    New role
```
JBuma: This should probably be a prop you can pass trough so this becomes more reusable.

src/views/RolesOverview.vue
```diff
@@ -9,10 +9,14 @@
             <span class="xr-title">Extinction Rebellion Nederland.</span>
           </h1>
         </div>
+        <new-item-dialog v-model="dialog" />
         <div v-if="$vuetify.breakpoint.smAndDown" class="mb-8">
           <v-divider />
-          <div class="d-flex justify-end pa-3">
-            <v-btn text color="primary" @click="drawer = true">Filter</v-btn>
+          <div class="d-flex justify-space-between pa-3">
+            <new-item-button @click="dialog = true" />
```
JBuma: Not hugely important, but you could extract this (`dialog = true`) into a method, since it gets used a few times in this file.

SamLubbers:
@JBuma thank you for your feedback. I just incorporated the changes you suggested along with the other functionality I still had to include.

For now, I have kept the email address as a required field and the phone number and matter most as optional because:
1. Not everyone has Mattermost. 
2. The validation is much easier.
3. It is easier to understand for the user (required email + optional contact fields) instead of (leave either an email or a phone number, and if you want also your Mattermost id)

For now, I have also left out the additional contact detail field so we keep the form as simple as possible. If our users think that our current fields are limited we can always add more options in the future.

A few more comments that will make some changes easier to understand:
* Reusable button: You mentioned making the `NewItemButton` more reusable by passing props. I have done that by creating the `IconButton`component. However, I have still kept `NewItemButton` because the same button is used in two places. That way, if we ever what to change that button (icon, text or colour) we just need to change the code in one location. 
* Responsibilities validation: as you can see, all fields are validated using VeeValidate. For the `responsibilities` field, I had to combine VeeValidate with some custom validation:
   * If there are 0 `responsibilities`, the `requiredList` VeeValidate rule shows an error message and prevents the form from being submitted. The user can still submit new responsibilities with this error.
   * If there are 10 responsibilities or `newResponsibility` has an invalid input, the `validResponsibility` computed property prevents users from adding adding to the list of `responsibilities` and the `errorResponsibility` computed property takes care of showing the error message. This is deliberately NOT a VeeValidate validator, because the user should still be able to submit the form if these conditions are not met (as long the user has entered at least 1 valid responsibility. Lastly, the `isEmpty` condition does not have an error message because so the user does not directly see an error when they open the form.

Issue #6, opened 2020-02-02
Task overview list


JBuma:
Users need to be able to see a list of tasks and filter them to find one they want to work on.
Basically the same thing as the existing role overview, but for tasks.
Filters needed:

- Text search (Search in description, title, location)
- Group filters (Local groups, working groups)
- Time commitment
- Location

Task cards have the following info, clicking on them will bring you to the task detail view:
- Title
- Local and working Groups
- Time commitment
- Button to view the detail page

SamLubbers:
To be completed after merging #12, when the overview and search of the role are linked to the backend.

SamLubbers:
We should minimise code duplication with the same functionality for roles. To achieve this we can: 1. Duplicate the existing role component
2. Edit the code to fit tasks.
3. Abstract into a base component the code that can be shared between tasks and roles.
4. Have the role and task component inherit from the base component.

abudaan:
I clone your fork and checkout the task-overview branch. Then when I run `npm i` and `npm run serve` it yields a massive amount of errors and warnings from prettier/prettier.

PR #31, opened 2020-02-22
Task overview

JBuma:
I got a bit carried away (as i tend to do), so this is kind of a big one. 

- Added a task overview page at /tasks, no links yet, so you'll have to type it in manually for now
- The list uses the same filters from the role overview, they're similar now, but they may diverge at some point. I just copy-pasted it for now, no need for unnecessary abstractions yet.
- Since this is very similar to the Role overview, i extracted a few components that were fairly reusable:
  - A default card, with title, subtitle, content, and action slots
  - A default drawer, with title slot and a close button
  - A page with drawer, which should take care of rendering the drawer, just pass the `isDrawerOpen` prop
  - A grid list, which uses css grid, to make the cards line up nicely

I'm not sure if i like the interface for the DefaultCard, if anyone has some opinions on this, please let me know.
Solves #6

### DIFF
--- a/src/components/layout/DefaultDrawer.vue
+++ b/src/components/layout/DefaultDrawer.vue
@@ -31,50 +31,13 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-.drawer {
-  position: fixed;
-  top: 0;
-  right: 0;
-  bottom: 0;
-  height: 100%;
-  width: 100%;
-  border-left-style: solid;
-  border-left-width: 1px;
-  z-index: 16;
-  transition: transform 0.3s ease-out;
-  overflow-y: auto;
-  transform: translateX(100%);
-  &.active {
-    transform: translateX(0);
-  }
-  .theme--light & {
-    background: white;
-    border-color: rgba(0, 0, 0, 0.12);
-  }
-  .theme--dark & {
-    background: #121212;
-    border-color: rgba(255, 255, 255, 0.12);
-  }
-
-  .drawer-header {
-    display: flex;
-    align-items: center;
-    border-bottom: 1px solid lightgray;
-    padding: 0.5rem;
-  }
-  .drawer-content {
-    padding: 0.5rem;
-  }
+.drawer-header {
+  display: flex;
+  align-items: center;
+  border-bottom: 1px solid lightgray;
+  padding: 0.5rem;
 }
-
-.bottom-border {
-  border-bottom-style: solid;
-  border-bottom-width: 1px;
-  .theme--light & {
-    border-color: rgba(0, 0, 0, 0.12);
-  }
-  .theme--dark & {
-    border-color: rgba(255, 255, 255, 0.12);
-  }
+.drawer-content {
+  padding: 0.5rem;
 }
 </style>
--- a/src/components/layout/DefaultDrawer.vue
+++ b/src/components/layout/DefaultDrawer.vue
@@ -10,7 +10,7 @@
       <v-btn
         v-if="this.$vuetify.breakpoint.smAndDown"
         icon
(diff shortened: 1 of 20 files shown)


abudaan:
If I clone your fork and checkout the task-overview branch and then run `npm i` and `npm run serve` it yields a massive amount of errors and warnings from prettier/prettier. Which is weird because I assume that you've committed your code with a code style that matches the settings in package.json

My work-around is to change the package.json settings to:
```
    "rules": {
      "prettier/prettier": "warn",
```
But off course that doesn't enforce the proper code style.

src/components/roles/RoleCard.vue
```diff
@@ -67,38 +60,9 @@ export default {
...
-  opacity: 0.7;
-}
-
-.transition-wrapper {
-  transition: background-color 280ms;
-}
-
-.lighter {
-  background-color: #333333 !important;
-}
```
SamLubbers: Include these classes back in for elevation of cards in dark mode.

src/components/FilterDrawer.vue
```diff
@@ -1,165 +0,0 @@
...
-.bottom-border {
-  border-bottom-style: solid;
-  border-bottom-width: 1px;
-  .theme--light & {
-    border-color: rgba(0, 0, 0, 0.12);
-  }
-  .theme--dark & {
-    border-color: rgba(255, 255, 255, 0.12);
-  }
-}
```
SamLubbers: add dark mode classes back in.

src/components/tasks/TaskFilters.vue
```diff
@@ -0,0 +1,79 @@
+<template>
+  <!-- Right now, this is the exact same as RoleFilters.vue, it could be that they will diverge at some point though, so for now there's two seperate files -->
+  <div>
+    <div>
+      <v-text-field
+        :value="selectedFilters.text"
+        label="Facilitator, Writer, Photographer..."
```
SamLubbers: replace with example tasks.

src/components/layout/DefaultDrawer.vue
```diff
@@ -0,0 +1,43 @@
+<template>
+  <div>
+    <div
+      v-if="
+        (!!$slots.header && !!$slots.header[0]) ||
+          this.$vuetify.breakpoint.smAndDown
```
SamLubbers: The header should only show if `this.$vuetify.breakpoint.smAndDown == true`, however, this expression would still be true if `(!!$slots.header && !!$slots.header[0]) == true`. I think this can be fixed by replacing the outer `||` with `&&`

src/components/FilterDrawer.vue
```diff
@@ -1,165 +0,0 @@
...
-      class="d-flex justify-space-between align-center pa-3 bottom-border"
-    >
-      <div class="d-flex align-center">
-        <v-btn icon @click="$emit('input', false)">
-          <v-icon color="primary">mdi-arrow-left</v-icon>
-        </v-btn>
-        <span>
-          <strong class="primary--text">{{ roleAmount }}</strong>
-          positions found
-        </span>
```
SamLubbers: It would be useful to show how many roles or tasks are found. Otherwise, when you are filtering on mobile (where the filter spans the entire page) you have no idea how many roles there are with each filter.

src/components/layout/DefaultDrawer.vue
JBuma: This used to be the case, but i moved the `Search for positions` and the clear filters button in there as well, just to make some space. So now it does need to show when: there is a header slot being passed, or when the screen is small enough to show the closing button. 
I could revert it back, but i think it looks just a little cleaner on mobile.

Issue #23, opened 2020-02-16
Edit role


SamLubbers:
Implement use case [#10: Editing a role](https://github.com/xrnl/aginun/wiki/Functional-Specification#10-editing-a-role), except for the steps that require authentication.

Resolve after completing #1 so we can reuse the interface for creating a new role & the code for querying the backend.

SamLubbers:
Button implemented, now we just need to hook it to the backend.

Issue #21, opened 2020-02-16
Delete role


SamLubbers:
1. Add delete button in the role detail component to delete a role
2. When the user clicks the delete button a prompt should ask the user if it wishes to confirm the deletion.
3. When the user clicks ok the role is deleted from the database and removed from the role overview page.

SamLubbers:
This issue was resolved in #51 🎉

Issue #19, opened 2020-02-16
Mark role as filled


SamLubbers:
Completes [use case 6.3](https://github.com/xrnl/aginun/wiki/Functional-Specification#6-applying-for-a-role)

Subtasks:

1. Button on the role dialog to mark a role as filled
2. When the button is clicked, make a query to add a timestamp to the role in the database marking it as filled
3. If the query is successful, show feedback to the user through the `alert.js` module in the Vuex store (as is done with creating, editing or deleting a role)
3. Only show roles that are not filled.
4. Indicate on the individual role view that it is filled. This is necessary if people navigate directly to a role through the URL.

Issue #7, opened 2020-02-02
Task detail view


JBuma:
Clicking on a task in the overview will bring a user to the detail view. This is preferably in a dialog, but the url should still change so people can copy the link and share it with others.

This functionality already exists for roles, so we can reuse most of the existing code. We just need to first resolve:
1. #15 to improve the role dialog
2. #6 for the task overview list.

PR #35, opened 2020-02-28
Role view improvements

### LINKED ISSUE #15: Improve role details dialog
Here are a few minor changes to improve the viewing the details of an individual role.

#### Content
* Display responsibilities as a list of items (one item for each responsibility), rather than a paragraph of text. 
* Mention how to apply for the role (see specifcation 6.1)

#### Navigation
* small button to close the role dialog
* open role dialog by clicking anywhere on the role card, rather than just 'read more'

#### Responsive design
* show all information into one column on small screens. Currently, the role description text is a bit squished.

SamLubbers:
Before making the task detail view (#7) it is a good idea to finalise the design for the role view. Then we can just copy the design for tasks. I made the following improvements to the role view:

1. Removed sidebar. There was too little content in the sidebar and too much content in the main text.
2. Added at the bottom of the description the time commitment information that used to be in the sidebar.
3. Added missing role fields (published date, requirements & responsibilities)
4. Added button to apply for the role. When this is clicked a modal shows with the contact details, each of which can be clicked. I acknowledge that having a modal inside of a modal is not an elegant solution, but it works fine for now. Also, I decided to add an "apply" button rather than directly adding the contact details for 2 reasons:
   4.1 I couldn't find a good place to add the contact details. Having the contact details on top didn't is too much because most people want to read the requirements/description first, having it on the bottom made it too hard to find and having them on the side like before created an imbalance of content (2.1)
   4.2 Another argument for having a button is that such a button will be needed when users apply for roles through the platform. If we have text now and later a button it can be confusing for users.
5. Added buttons for marking a role as taken, and buttons for edit and delete (first part of #19, #21 & #23). Note that these buttons will only show up for the accounts that create those roles. 

Lastly, I removed the nested CSS classes because they are not necessary in Vue components. The `scoped` attribute on the `style` tag ensures those classes are only available in that component. 

I would like to suggest that this PR closes #15, even though not all parts of that issue are completed. I will make separate individual issues for the parts that are not implemented yet.


JBuma:
I like how this looks, only thing i would change is move the `apply` and `role taken` buttons to a footer in the bottom, the other menu button fits okay there. People are more used to those action buttons being in the bottom right, so it would be more conventional that way. Also, if you use tab to select the buttons, they're in kind of a weird order (the apply buttons gets focus first, then the role taken button, then the menu button).

SamLubbers:
I added the buttons on top to make them easier to find but they indeed do not fit well there. Just added the buttons at the bottom and the tab issue is resolved.

Issue #37, opened 2020-02-29
solid color on card text


SamLubbers:
By default on vuetify, the text inside card descriptions (inside `v-card-text`) is lighter. See images below.

Light             |  Dark
:-------------------------:|:-------------------------:
<img width="300" alt="Screenshot 2020-02-29 at 15 26 17" src="https://user-images.githubusercontent.com/15846595/75609342-f1ed9780-5b07-11ea-81c3-792642e3bbb9.png"> | <img width="300" alt="Screenshot 2020-02-29 at 15 26 43" src="https://user-images.githubusercontent.com/15846595/75609344-f2862e00-5b07-11ea-9732-61ca462d9e1f.png">

We want this text to have the regular solid font colour. The way we take care of this right now is to use CSS classes in each component that uses `v-card`. This is clearly inefficient as we would have to use these classes every time we use `v-card`. 

To set the font colour for card-text we can override Vuetify globals. The global that makes this text lighter is `'secondary-text-percent'` in `$material-dark` & `$material-light` (found in `vuetify/src/styles/settings/_dark.scss` and `vuetify/src/styles/settings/_light.scss`.

In theory, we should be able to override this global by adding the following to `src/styles/variables.scss`

```scss
$material-light: (
  "secondary-text-percent": map-get($material-light, "primary-text-percent"),
);

$material-dark: (
  "secondary-text-percent": map-get($material-dark, "primary-text-percent"),
);
```
This does not work at the moment, but it should in theory work because there are other globals that we override with this method

```scss
$material-light: (
  "background": map-get($shades, "white"),
);

$material-dark: (
  "background": #121212,
    "cards": #1e1e1e,
);
```

Issue #36, opened 2020-02-28
View by clicking card


SamLubbers:
open the detail view of a role or task by clicking anywhere on the card in the overview, rather than just 'read more'. The cursor should also become a pointer when hovering over those cards.

Issue #33, opened 2020-02-23
Due date


SamLubbers:
#### Roles
When creating new roles, rebels should specify until which date the role position is relevant. When that date is passed and the role is still not filled, volunteers shouldn't be able to see that role anymore. This solves two problems:

* Volunteers applying for outdated positions and getting a rejection or no reply at all. If this happens once it is okay but if it happens multiple times volunteers will lose trust in our system and stop using it. 
* Not being able to find relevant roles and positions.

#### creation
1. see due date in the role detail view
2. add due date when creating a new role

abudaan:
I have added the fields `date_added` and `due_date` to the role and task tables.

SamLubbers:
Closed in #99 :tada:

PR #42, opened 2020-03-08
Task detail view

### LINKED ISSUE #5: Task creation
Implement all steps for the use case [#3: creating a task](https://github.com/xrnl/aginun/wiki/Functional-Specification#3-creating-a-task), except for the steps that require user authentication.

This issue can be easily completed once #1 is done.

JBuma:
I added a task creation dialog, with some store mutations, and the necessary routes. It is absolutely not DRY, since we want to get this in the air as quickly as possible, i copy-pasted a lot from the roles. Once we have a little more time, this will need to be refactored.
Also fixed some smaller issues: #37, and #36 

 closes #5 #7 #37, and #36

### DIFF
--- a/src/components/roles/RoleCard.vue
+++ b/src/components/roles/RoleCard.vue
@@ -1,5 +1,5 @@
 <template>
-  <default-card>
+  <default-card :to="`roles/view/${role.id}`">
     <template #header>
       {{ role.workingGroup.text }}
     </template>
@@ -17,11 +17,6 @@
         </span>
         <span class="overline text-uppercase">hours / week</span>
       </span>
-    </template>
-    <template #action>
-      <v-btn text dark color="primary" :to="`roles/view/${role.id}`">
-        Read More
-      </v-btn>
     </template>
   </default-card>
 </template>
--- a/src/components/roles/RoleCard.vue
+++ b/src/components/roles/RoleCard.vue
@@ -18,6 +18,11 @@
         <span class="overline text-uppercase">hours / week</span>
       </span>
     </template>
+    <template #action>
+      <v-btn text dark color="primary" :to="`roles/view/${role.id}`">
+        Read More
+      </v-btn>
+    </template>
   </default-card>
 </template>
 
--- a/src/components/tasks/NewTaskDialog.vue
+++ b/src/components/tasks/NewTaskDialog.vue
@@ -1,9 +1,9 @@
 <template>
   <v-dialog :value="value" width="600px" @input="$emit('input', false)">
     <v-card class="pa-4">
-      <h2>New role</h2>
+      <h2>New Task</h2>
       <validation-observer ref="form" v-slot="{ invalid, handleSubmit }">
-        <form @submit.prevent="handleSubmit(publishRole)">
+        <form @submit.prevent="handleSubmit(publishTask)">
           <validation-provider
             v-slot="{ errors }"
             rules="required|alpha_spaces|max:30"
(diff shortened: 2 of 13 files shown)


src/components/roles/RoleCard.vue
```diff
@@ -18,11 +18,6 @@
         <span class="overline text-uppercase">hours / week</span>
       </span>
     </template>
-    <template #action>
-      <v-btn text dark color="primary" :to="`roles/view/${role.id}`">
-        Read More
-      </v-btn>
-    </template>
```
SamLubbers: can you add the 'read more' text back on the card? Even if the entire card is clickable it is still useful that there is some text indicating that users can click on the card to read more. See [this interface](https://careers.google.com/jobs/results/?company=Google&company=YouTube&company=Google%20Fiber&employment_type=FULL_TIME&hl=en_US&jlo=en_US&q=&sort_by=relevance) as an example. It has the `expand` text at the bottom of each card, even though it is possible to expand by clicking anywhere on the card.

src/components/tasks/NewTaskDialog.vue
```diff
@@ -0,0 +1,291 @@
...
+              item-value="value"
+              item-text="text"
+              return-object
+              label="Working group"
+              :error-messages="errors"
+            />
+          </validation-provider>
+          <p style="color: gray" class="caption">
+            Responsibilities
+          </p>
```
SamLubbers: remove responsibilities as that was specific to only roles.

src/components/tasks/NewTaskDialog.vue
```diff
@@ -0,0 +1,291 @@
...
+            <v-textarea
+              v-model="requirements"
+              label="Requirements (optional)"
+              placeholder="Skills, experience, equipment"
+              :error-messages="errors"
+            />
+          </validation-provider>
+          <validation-provider
+            v-slot="{ errors }"
+            rules="max:1000"
```
SamLubbers: The task description should be required as tasks do not have the `responsibilities` field.
```suggestion
            rules="required|max:1000"
```

src/components/tasks/NewTaskDialog.vue
```diff
@@ -0,0 +1,291 @@
...
+            name="description"
+          >
+            <v-textarea
+              v-model="description"
+              label="Description (optional)"
```
SamLubbers: description should be required (see previous comment)
```suggestion
              label="Description"
```

src/components/tasks/NewTaskDialog.vue
```diff
@@ -0,0 +1,291 @@
...
+              placeholder="Any additional information not specified in the set of responsibilities.
+        
+This can include information about the circle or the specific project that the task is needed for."
```
SamLubbers: Updated the placeholder because tasks do not have responsibilities & the explanation of the task needs to go in the description. 
```suggestion
              placeholder="Purpose of the task and how to complete it.  
              
You can also include other relevant information such as the circle or the project that the task is needed for."
```

src/store/modules/meta.js
```diff
@@ -6,5 +6,12 @@ export default {
       { min: 11, max: 20 },
       { min: 21, max: 30 },
     ],
+    taskTimeCommitments: [
+      { min: 1, max: 5 },
+      { min: 6, max: 10 },
+      { min: 11, max: 20 },
+      { min: 21, max: 30 },
+      { min: "More" },
```
SamLubbers: I suggest removing `more` as an option for time commitment, as a task of more than 30 hours can be too long and could possibly be split into smaller tasks. 

If you think we should still leave the option I suggest you replace the word `more` with `30+` That way users see `30+ hours` on the card instead of `more hours`. 
```suggestion
```

src/components/tasks/TaskCard.vue
```diff
@@ -15,11 +15,6 @@
         <span class="overline text-uppercase">hours</span>
       </span>
     </template>
-    <template #action>
-      <v-btn text dark color="primary">
-        Read More
-      </v-btn>
-    </template>
```
SamLubbers: add `read more` text back into task card. (see comment in `RoleCard.vue`)

JBuma:
Incorporated your requested changes. I haven't done #33 yet, that might be better to do in a different pr.

PR #50, opened 2020-04-12
Setup unit testing

SamLubbers:
Wrote tests for the first few components in `src/components/__tests__/components.spec.js `, as specified in the. I configured tests as specified by [vue-test-utils](https://vue-test-utils.vuejs.org/) and [Vuetify testing](https://vuetifyjs.com/en/getting-started/unit-testing/)


SamLubbers:
@moritzcopter this pull request is FINALLY ready ✊. I have made unit tests for about half of the components in `src/components`. You can find all the tests in `src/components/__tests__/components.spec.js`. 

The commit history is a bit of a mess, so I suggest directly reviewing the final code. I think the best way to go about it is to review one component at a time: start in `components.spec.js` and find the first `describe` block, which contains all the tests for one component. Then review the changes for that component (in `src/components/*.vue`) and then the tests for that component (in the `describe` block).

Please let me know if I have defined the right tests for each component. In component unit testing, like other unit testing, the goal is to verify that you get the expected outputs for certain inputs. When testing vue components, the inputs and outputs are the following:
  * inputs:
    * component props & slots
    * user interactions (e.g. button click)
    * Global data (from the Vuex store) that determines the behaviour of the component. For example, the type of user that is logged in. 
  * outputs:
    * rendered HTML
    * emitted events (e.g. a vuex mutation is called when user clicks on a button)

If you need help understanding the syntax, you can check out the following resources:
  * [Vue test utils](https://vue-test-utils.vuejs.org/)
  * [Jest](https://jestjs.io/docs/en/getting-started.html)
  * [Vuetify testing](https://vuetifyjs.com/en/getting-started/unit-testing/)

Thank you for reviewing this pull request 💚

SamLubbers:
> What do you think about adding a Check in github that runs the tests? Is that easily possible?

@moritzcopter yes that's certainly possible with [required status checks](https://docs.github.com/en/github/administering-a-repository/about-required-status-checks). It's one of the next things I'm going to implement 👌

SamLubbers:
@moritzcopter just merged this PR 🎉  Resolving the merge conflicts took a while but it's done and all tests are working on the master branch. Now you can also write tests for components if you want 😉 If that's too boring I can continue taking care of the tests and you can resolve some other issues 👌

PR #70, opened 2020-06-27
Do not create/edit role/task on enter

### LINKED ISSUE #53: do not create/edit role on enter
When the user presses `enter` in the dialog to create or edit a role, the action from the `submit` button is executed. We do not want this behaviour because they might press enter when filling in one of the fields, and inadvertently create or edit the role too early.

moritzcopter:

Override the default behavior of submitting the create / edit role forms by pressing enter

Resolves #53 

## Additional comments

See https://stackoverflow.com/questions/44261073/vuejs-how-to-prevent-form-submit-on-enter-but-still-save-work-in-inputs

### DIFF
--- a/src/components/roles/NewRoleDialog.vue
+++ b/src/components/roles/NewRoleDialog.vue
@@ -5,7 +5,7 @@
       <validation-observer ref="form" v-slot="{ invalid, handleSubmit }">
         <form
           @submit.prevent="handleSubmit(publishRole)"
-          @keydown.enter.prevent
+          @keypress.enter.prevent
         >
           <validation-provider
             v-slot="{ errors }"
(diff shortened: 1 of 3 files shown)


src/components/roles/NewRoleDialog.vue
```diff
@@ -3,7 +3,10 @@
     <v-card class="pa-4">
       <h2>New role</h2>
       <validation-observer ref="form" v-slot="{ invalid, handleSubmit }">
-        <form @submit.prevent="handleSubmit(publishRole)">
+        <form
+          @submit.prevent="handleSubmit(publishRole)"
+          @keydown.enter.prevent
```
SamLubbers, line 8: This solution fixes issue #53, but now users cannot add role responsibilities anymore by pressing enter. This is easily fixed by replacing `@keydown` with `@keypress`, because `@keypress` is the command used (in line 63) to add new responsibilities.
```suggestion
          @keypress.enter.prevent
```

PR #73, opened 2020-07-05
Refactor role creation & edit

### LINKED ISSUE #66: Refactor role creation & edit
Currently there is a lot of duplicate code between `newRoleDialog.vue` and `editRoleDialog.vue`. As much code as possible should be reused between these two components so that the code is easier to maintain.

moritzcopter:

Reduce duplicate code between RoleEditDialog and NewRoleDialog by using a [Mixin](https://vuejs.org/v2/guide/mixins.html).
Additionally extract role properties of the component form (commit https://github.com/xrnl/aginun/commit/944dcc5f74b2c3646ed3eb363e09bb2bab66e856) into a separate object, which makes it easier to create/edit roles and separates role properties and the other component properties more cleanly

Resolves #66


src/components/roles/NewRoleDialog.vue
```diff
@@ -11,8 +11,7 @@ export default {
   methods: {
     ...mapActions("roles", ["createRole"]),
     onSubmit: function() {
-      const role = JSON.parse(JSON.stringify(this.$data.role));
-      this.createRole(role);
+      this.createRole(this.role);
```
moritzcopter, line 14: @SamLubbers Can we remove this parsing back-and-forth or am I missing something? Works for me fine without..

SamLubbers: Yes that is some confusing code 🤔 . It is [used for cloning an object](https://github.com/vuejs/vue/issues/158#issuecomment-158770833). It would be good to add a comment or change the variable name to `role_clone`.

What that code does is create a clone of the role object, remove properties from the role (clone) object that we don't want to store in the database (`newResponsibility` and `$apolloData`), and call `createRole` to store the role data in the database. If I remember well, without making the clone, the code would also remove the properties from the data object in the component, meaning that it wouldn't be possible to add responsibilities when creating another new role. 

I have opted for this solution (of removing properties we don't want to send) because the alternative has a lot of repetitive code:

```js
const role_data_to_store_in_db {
  title: this.role.title,
  description: this.role.description,  
  requirements: this.role.requirements,
  responsibilities: this.role.responsibilities,
  // repeat for each of the properties...
}
```

moritzcopter:
@SamLubbers Note that I just pushed another commit after realising that it might be easier to just use a single component for edit and create instead of a mixin - what do you think?

src/components/roles/NewRoleDialog.vue
SamLubbers: @moritzcopter sorry, I replied to your comment before looking at all your code changes 😓. Now that the role data is in a separate `role` object instead of in `this.$data`, I think we can refactor the code the way you did it. I'll finish my review and then confirm.

moritzcopter:
> 1. change the `role` properties (in `initialState`) to match the structure of the role loaded from the database:
> 
> * `timeCommitment.min` and `timeCommitment.max` instead of `timeCommitmentMin` and `timeCommitmentMax`
> * `localGroup.id` and `workingCircle.id`instead of `localGroupId` and `workingCircleId`

But then we also need to change the role format in the `CreateRoleMutation` and `UpdateRoleMutation` endpoints no?

SamLubbers:
> But then we also need to change the role format in the `CreateRoleMutation` and `UpdateRoleMutation` endpoints no?

@moritzcopter you are right. I did not realize that the fields of the `role` object in `RoleEditDialog` match the fields of the input object for creating and editing a role.

src/components/roles/RoleEditDialog.vue
```diff
@@ -284,18 +284,25 @@ export default {
...
+        if (key in this.editRole) {
+          this.role[key] = this.editRole[key];
+        }
+      }
+      this.role.workingCircleId = this.editRole.workingCircle.id;
+      this.role.localGroupId = this.editRole.localGroup.id;
+      this.form_title = "Edit Role";
+    } else {
+      this.form_title = "New Role";
+    }
```
SamLubbers, line 305:
```suggestion
```

src/components/roles/RoleEditDialog.vue
```diff
@@ -1,18 +1,16 @@
 <template>
-  <!-- TODO: This is very similar (almost the same) as the create, at this point it would be nice if we could get a generic solution for this.
-        Will have to look into this once we will implement the tasks.-->
   <v-dialog :value="value" width="600px" @input="$emit('input', false)">
     <v-card class="pa-4">
-      <h2>Edit role</h2>
+      <h2>{{ form_title }}</h2>
```
SamLubbers, line 4:
```suggestion
      <h2>{{ formTitle }}</h2>
```

src/components/roles/RoleEditDialog.vue
```diff
@@ -318,29 +325,28 @@ export default {
     },
```
SamLubbers, line 325:
```suggestion
    },
    formTitle: function() {
      return this.editRole ? "Edit Role" : "New Role";
    },
```

src/components/roles/RoleEditDialog.vue
```diff
@@ -318,29 +325,28 @@ export default {
...
   },
```
SamLubbers, line 326:
```suggestion
  },
  watch: {
    editRole: {
      handler: function(editRole) {
        if (editRole) {
          for (var key in this.role) {
            if (key in editRole) {
              this.role[key] = editRole[key];
            }
          }
          this.role.workingCircleId = this.editRole.workingCircle.id;
          this.role.localGroupId = this.editRole.localGroup.id;
        }
      },
      immediate: true,
    },
  },
```

moritzcopter:
Hi Sam, thanks again for taking the time! Unfortunately I only commited your proposed changes in the browser but only verified now that they work locally - looks like there was a small bug introduced which should be fixed by: https://github.com/xrnl/aginun/pull/75

PR #88, opened 2020-10-25
easy clearing of filter options

### LINKED ISSUE #55: easier clearing of filters
The following improvements will make it easier for the user to clear filters, and therefore to find roles.

### Search
Currently, when a user wants to remove their search term they need to eliminate all characters one by one.

An easier approach, is that when a user types text into the search bar, a small cross icon appears in the search field. When the user clicks that icon, all text is cleared and the icon disappears.

### Local group and working circle filters
Currently, when a user wants to remove a selected working group or local group, they need to click the input field, scroll down the list of groups and deselect the desired one. 

An easier approach, is that when a user selects a local group or working circle, a small cross icon appears next to the working group or local group name. See an example [here](https://vuetifyjs.com/en/components/combobox/#advanced-custom-options). When a user clicks on the cross, the group should be deselected.

kajdehoop:

Correct props have been added to the Vuetify elements. RoleFilters input can now be cleared with cross, local group and working circle field selections can be removed with cross.

**Before:** 
before

**After:**
after

Resolves #55 

## Additional comments

The clear button on the RoleFilters sets the field to null when clicked. The filtering does not work when the input is null, only when it's "" (empty string). I added a check on the filterValue that sets it to "" if it's null in order to fix this issue.

### DIFF
--- a/src/components/AutocompleteCustom.vue
+++ b/src/components/AutocompleteCustom.vue
@@ -6,6 +6,7 @@
     item-text="title"
     chips
     small-chips
+    deletable-chips
     multiple
     class="mt-3"
     :label="label"
--- a/src/components/roles/RoleFilters.vue
+++ b/src/components/roles/RoleFilters.vue
@@ -7,6 +7,7 @@
         placeholder="Facilitator, Writer, Photographer..."
         class="mt-3"
         @input="debounceSearchUpdate"
+        clearable
       />
     </div>
     <filter-section>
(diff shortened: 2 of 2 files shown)


SamLubbers (approved):
@kajdehoop thank you for implementing this useful improvement 🙌  It works exactly like expected so I'll merge is straight away in the main codebase 💯 

Congratulations with your first contribution to this project 🎉  It's awesome to have you in the team 😎

kajdehoop:
Blij dat ik kan helpen 😄

PR #101, opened 2020-11-02
fixed padding on role cards and views

### LINKED ISSUE #92: Add padding to popups
The popup modal feel a little cramped right now

## Feature description
Add some padding to the popups, so they look better

kajdehoop:

Added 10px of padding to the role cards and views. Made role cards 20px higher (now 200px)

**Before:**
before1
before2

**After:**
after1
after2

I added padding to make the cards and views feel less cramped. The role cards have been made 20px higher to account for the extra padding.

Resolves #92

### DIFF
--- a/src/components/roles/RoleCard.vue
+++ b/src/components/roles/RoleCard.vue
@@ -37,7 +37,7 @@ export default {
   components: {
     DefaultCard,
   },
-    props: {
+  props: {
     role: {
       type: Object,
       required: true,
@@ -54,4 +54,8 @@ export default {
 };
 </script>
 
-<style lang="scss" scoped></style>
+<style lang="scss" scoped>
+.v-card {
+  padding: 10px;
+}
+</style>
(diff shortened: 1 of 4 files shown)


src/components/roles/RoleCard.vue
```diff
@@ -54,4 +54,8 @@ export default {
 };
 </script>
 
-<style lang="scss" scoped></style>
+<style lang="scss" scoped>
+.v-card {
+  padding: 10px;
+}
+</style>
```
SamLubbers, line 61: I would move this to `DeafaultCard.vue`. That way, when we make cards for tasks they will also have the increased padding.

SamLubbers (changes requested):
@kajdehoop thanks for this change and sharing very clear screenshots :raised_hands: it does indeed look less cramped with this padding. I've just added a small suggestion in the code but after resolving that I can merge it.

SamLubbers:
Closing this PR as it's quite old and not merged into the master branch.

PR #110, opened 2020-11-08
Server error

### LINKED ISSUE #106: Handle server connection errors

1. remove the API key from `.env.local`
2. start the application

Friendly error message explaining that the connection to the server has failed and that they can email tech@extinctionrebellion.nl for help.

If the application is in development mode, it should also suggest to the user reading the `README.md` and properly setting up the server API key.

These error messages should also appear if the user directly goes to the view of an individual role.

## Actual result

Loading... Loading... Loading...

aleesaan:
Added a text to show to the user when there was an error fetching roles.

Resolves #106 

## Additional comments
We should also add error tracking, to notify us if these errors happen on production, but it will be done on a different case.

### DIFF
--- a/src/views/ErrorPage.vue
+++ b/src/views/ErrorPage.vue
@@ -7,10 +7,7 @@
       Something went wrong 😧
       <br />
       Please email
-      <a href="mailto:tech@extinctionrebellion.nl">
-        <!-- eslint-disable-next-line vue/html-closing-bracket-newline -->
-        tech@extinctionrebellion.nl</a
-      >
+      <a :href="`mailto:${contactEmail}`">{{ contactEmail }}</a>
       for help.
     </p>
     <p v-else>
@@ -22,17 +19,18 @@
       .
       <br />
       If the problem persists please contact
-      <a href="mailto:tech@extinctionrebellion.nl">
-        tech@extinctionrebellion.nl
-      </a>
+      <a :href="`mailto:${contactEmail}`">{{ contactEmail }}</a>
     </p>
   </div>
 </template>
 
 <script>
+import { contactEmail } from "@/constants/contacts";
+
 export default {
   name: "ErrorPage",
   data: () => ({
+    contactEmail,
     isDevMode: process.env.NODE_ENV !== "production",
   }),
 };
--- a/src/views/ErrorPage.vue
+++ b/src/views/ErrorPage.vue
@@ -33,7 +33,7 @@
 export default {
   name: "ErrorPage",
   data: () => ({
-    isDevMode: process.env.NODE_ENV === "production",
+    isDevMode: process.env.NODE_ENV !== "production",
   }),
 };
 </script>
--- a/src/views/ErrorPage.vue
+++ b/src/views/ErrorPage.vue
@@ -1,0 +1,48 @@
+<template>
+  <div class="pa-6 text-center">
+    <h1 class="text-sm-large">
+      DON'T PANIC.
+    </h1>
+    <p v-if="!isDevMode">
+      Something went wrong 😧
+      <br />
+      Please email
+      <a href="mailto:tech@extinctionrebellion.nl">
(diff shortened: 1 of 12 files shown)


SamLubbers (changes requested):
@aleesaan thanks for picking up this issue :raised_hands: 

I would slightly change the implementation so that it handles the server error no matter what page the user visits. Right now the user sees the error message when visiting the `RolesOverview`, but not when they go to a different page, such as viewing an individual role:

image

I think the easiest way to deal with this error is to add the error message in `App.vue`: if the connection to the server fails show a page with only the error message, else, render all the components of the application everything that is in `v-app` at the moment. I would show this error message big and centred, something like this :stuck_out_tongue_winking_eye: 

image
 
I would also reuse the support contact details in `TheAppBar`. That way if we change these contact details it automatically changes in all places. The easiest way to do this is to make a shared component used in both places. As you can see in `TheApprBar`, we still use the integration@extinctionrebellion.nl email. You can change that with tech@extinctionrebellion.nl

As a last small detail, I would make the text `set the server API key` a hyperlink to: https://github.com/xrnl/aginun#install That way developers know how they can set the API key.

src/views/ErrorPage.vue
```diff
@@ -0,0 +1,48 @@
...
+      </a>
+    </p>
+  </div>
+</template>
+
+<script>
+export default {
+  name: "ErrorPage",
+  data: () => ({
+    isDevMode: process.env.NODE_ENV === "production",
```
SamLubbers, line 36: small typo here :wink: 
```suggestion
    isDevMode: process.env.NODE_ENV !== "production",
```

SamLubbers (changes requested):
Hey @aleesaan thanks for implementing the feedback :smile: This error page is already so much better than the previous version, as the user cannot interact with the application when the connection to the server fails.

### More resilient error handling

I think, however, that we can make the application even more robust. Right now the error page shows whenever the user goes to one of the roles pages (`/roles`) and the application fails to load roles. This works fine for now, but once we add more pages (`/tasks`, `/settings`) the error page will not load even if the connection with the server fails. That is why I suggest adding the error handling within a [beforeCreate lifecycle hook](https://vuejs.org/v2/api/#beforeCreate) in `App.vue`. That way, if the connection to the server fails, the application will not even attempt to load the roles.

In summary, the current flow is as follows:
  1. Application loads
  2. Application tries to load roles
  3. There is an error connecting to the server --> error page is shown

The new flow would be as follows:
  1. Application loads
  2. Inside the `beforeCreate` lifecycle hook of App.vue, the application tries to ping the server to see if a connection can be established.
  3. There is an error connecting to the server --> error page is shown

### Shared contact details

Right now the email `tech@extinctionrebellion.nl` is hardcoded into `TheAppBar.vue` and `ErrorPage.vue`. I suggest exporting the email into a variable (`contactEmail`), so that if this email changes it is automatically updated in all parts of the application. 

So you could create a file with all application constants:

```js
export const contactEmail = 'tech@extinctionrebellion.nl'
```
and then import this email in all places that you use it

```html
<a :href=`mailto:${contactEmail}`>
   {{ contactEmail }}
</a>
```

src/views/ErrorPage.vue
aleesaan: ahah thanks!

aleesaan:
Your suggestion sounds good to me! But wouldn't we need to also to configure an endpoint in the backend so that it can be pinged? Using one of the existent queries would be kind of random...

SamLubbers:
@aleesaan I had assumed that Hasura would have a default endpoint to ping. If it doesn't have it then we can indeed set up a custom one :ok_hand:

SamLubbers (approved):
@aleesaan this is PERFECT :star_struck: now this functionality is even better as the error page loads directly (instead of the roles page loading for a fraction of a second and then the error page showing). Also this saves us a lot of code checking for server errors in other parts of the application.

PR #112, opened 2020-11-08
Add utils tests

### LINKED ISSUE #62: Unit tests of utils
Unit tests for all files in `src > utils`.

aleesaan:
Added tests for utils functions.

Resolves #62


src/utils/__tests__/validators.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    expect(iconValidator("mdi-icon")).toBe(true);
+  });
+  it("should return false when the provided string doesn't start with 'mdi-'", () => {
+    expect(iconValidator("icon")).toBe(false);
+  });
+});
+
+describe("hrefValidator", () => {
+  it("should return true when the provided string starts with one of the valid prefixes", () => {
+    expect(hrefValidator("https://extinctionrebellion.nl")).toBe(true);
```
SamLubbers, line 56: To make the test more complete I would also add the validation for the other valid hyperlinks:

```js
  const validHrefPrefixes = ["http://", "https://", "mailto:", "tel:"];
```

SamLubbers (changes requested):
@aleesaan awesome that you've made tests for the utility functions :fist_raised: 

You've covered the most important test cases, and they all pass :100: I've just left one comment in the code about some cases that would make these tests more complete.

Besides that, I think we can simplify the test names:
  * Instead of saying what a test should do, make an assertion. So instead of `should return true` you can say `returns true`.
  * use less prepositions when possible: For example, instead of `should return true when all the items in the array pass the function provided` you can say `returns true when all array times pass validation function`
  * you can also remove some redundant words, so you can say `string` instead of `provided string` or just `structure` instead of `expected structure`

SamLubbers (approved):
@aleesaan Thanks for these last changes :star2:

PR #118, opened 2020-11-22
Added basic login logic

### LINKED ISSUE #41: Setup authentication
We need Authentication so that not everyone can create, edit, and delete roles & tasks. To limit these functionalities to local groups & working groups in XR, we will issue pre-created accounts to these groups. 

### subtasks
1. Chose an authentication service: We need such a service [authentication is handled outside of Hasura](https://hasura.io/docs/1.0/graphql/manual/auth/index.html)
2. Setup authentication service: Hasura supports [two authentication methods](https://hasura.io/docs/1.0/graphql/manual/auth/authentication/index.html): Webhooks or JWT tokens. We have been advised that: "if you want to get moving fast, the webhook looks much simpler. However, each requests received by Hasura will fire a request to the webhook as opposed to a single JWT token request which is then passed with each requesst. So that's probably something for later on when / if you see performance issues"

kajdehoop:

I've added a user.js Vuex module with login/logout functionality and added a login/logout button. The login function makes a request to the newly set up XR Keyserver. Currently we have a test user set up there ("kaj-dev" with password "test") that's hardcoded into the login button.  Upon clicking the login button, this user is authenticated by the Keyserver, and an access token (with a 1 day expiration) is retreived. This token is stored in the Vuex store. The loggedIn variable is also set to true. As a proof-of-concept, when the loggedIn is set to true the "New Role" button is shown, and when loggedIn is false, it's hidden.

**Before**

before

**After**

after 1
after 2

Resolves #41 

## Additional comments

2 more things are needed to provide a complete login functionality (for existing users), but I believe these should be separate issues:

- Login button should trigger a modal with 'username' and 'password' inputs, so the user can fill in his own credentials instead of the hardcoded ones that are currently used

- The token should be cached in the front-end, so a user stays logged in after refreshing the page. This can be done by setting a cookie (local storage) containing the token and invalidating this cookie after the expiration time of the token has passed or if a request is being made with an invalid token. The user should then be prompted to log in again.


src/store/modules/user.js
```diff
@@ -0,0 +1,49 @@
...
+        client_id: "volunteerplatform",
+        username,
+        password,
+      };
+
+      const auth = await axios.post(keyserverUrl, qs.stringify(params), config);
+
+      const token = auth.data.access_token;
+
+      console.log("Logged in! Access token: ", token);
```
SamLubbers, line 39: I assume you had this for testing. This can be removed as it works now :wink: 
```suggestion
```

src/store/modules/user.js
```diff
@@ -0,0 +1,49 @@
...
+
+      commit("setToken", token);
+      commit("setLoggedIn", true);
+    },
+    async logout({ commit }) {
+      commit("setToken", null);
+      commit("setLoggedIn", false);
```
SamLubbers, line 46: I see that whenever you log in and log out you set two variables: `token` and `loggedIn`. It might not seem like it, but this is a case of duplicate information, because whenever you set the token you also need to set the log in status. You can solve this by only setting the token, and then making a computed property for `loggedIn`

```js
  state: {
    token: "",
  },
  getters: {
    loggedIn: state => !!state.token,
  },
```  

Note that if you change this you'll also have to change `$store.state.user.loggedIn` with `$store.getters['user/loggedIn']`. Or even better, you can [mapGetters](https://vuex.vuejs.org/guide/modules.html#namespacing) so that you can directly use the variable loggedIn` within the HTML.

SamLubbers (changes requested):
@kajdehoop thank you so much for implementing this really important functionality that was missing from the platform :fist_raised: :tada: 

The log in and log out works flawlessly. Besides the feedback I left in the code, there is only one thing that I would change before merging this pull request:

### Hide other buttons

I see that you've already hidden the `new role` button for logged out users. There are a few more button that should not be shown to logged out users. In `components/roles/RoleViewDialog.vue`, we need to hide:
  * the button "Role filled"
  * the three dots on the top right which show the edit and delete buttons.

### Future functionality

I think it's a good idea to merge this core functionality and then add the remaining functionality later. Could you make issues for the points that you mentioned? You can assign them to the milestone `next verstion` and add the `auth` label so we can easily find all issues related to authentication.

### Authentication documentation

It would also be really useful if you add a page in the wiki explaining how the authentication system works: what systems it uses, where it is set up, how to create new users etc. That way, someone new joining the platform can understand how the Authentication system works and can help out.

This doesn't need to happen straight away. You can make an issue and get back to it later. Or maybe Felix could also help out :wink:

kajdehoop:
Good feedback! Fixed it all :) I've only left the login function in TheAppBar component as is (instead of mapping it directly to an action) because we need to replace this function with a function that opens a login modal in this place anyway. I'll make new issues for the login modal and the caching of the token, and add a small wiki page explaining the authentication. 

One more suggestion for something I saw in the code: I think it's best to remove all the name properties from our components (for example the name: "TheAppBar" property on TheAppBar), as this is duplicate information. The file name of the component already contains the component name and can be imported as such. If someone copies a component and changes the file name, they might forget to change the name property and that would cause conflicts with imports. By omitting the name property in components we prevent this problem from happening altogether.

SamLubbers (approved):
@kajdehoop thank you for implementing the feedback, making the issues and updating the documentation :grin: You can now merge this code into the master branch.

There are some merge conflicts with the master branch because I just merged #119, which has some changes to the ESLint settings. I think the easiest way to fix this is:
  * merge the master branch into your branch.
  * keep the master package.json & package-lock.json settings.
  * reinstall the packages that you installed in this PR (axios, vue-axios, qs)
  * keep your changes for the `.vue` files, and simply fix any linting issues you might have.

Lastly, I have made issue #123 for some more detailed authentication documentation :seedling:

kajdehoop:
@SamLubbers I needed to fix a bit more stuff than I had hoped, but now you should be able to merge it :)

SamLubbers (approved):

SamLubbers:
@kajdehoop in this PR you also mentioned this:

> I think it's best to remove all the name properties from our components (for example the name: "TheAppBar" property on TheAppBar), as this is duplicate information. 

I did forget to reply at the time but this hasn't disappeared from my todo list :wink: You're right that the name property is not needed and it adds some duplicate information. However, [as you can read here](https://vuejs.org/v2/api/#name), the name property is useful for debugging the code. For this reason I think it's useful to keep it

Issue #39, opened 2020-03-01
Dutch + English support


SamLubbers:
Right now English is the only supported language. We need to also support Dutch as XR NL is a Dutch movement. 

In addition to being able to change the text of the application, the user should also be able to:
* add roles/tasks in both dutch and english. This should be an option but should not be required. Users can still add roles/tasks in just dutch or just english.
* see roles/tasks in the language that they have selected
* if there are roles/tasks not written in the selected language, the user should get an indication of this when browsing.

kajdehoop:
We can use Vue i18n for translations of texts in the UI. We can combine this with Loco (https://localise.biz/) and get the updated translations from that API with gulp during the build process.

SamLubbers:
@kajdehoop that sounds good :ok_hand: I would also look into [Vuetify's i18n](https://vuetifyjs.com/en/features/internationalization/#getting-started) so translations work well with existing components.

In addition to translating fixed text, we also need to make sure that users provide all role information in both english and dutch (role description, responsibilities, requirements...). In addition, users should also be able to mark a role as only for dutch speakers. This role should then only appear in the dutch page but not in the English page.

SamLubbers:
@aleesaan you had experience with translations right? Would you like to work on this issue?

aleesaan:
I could give it a try, but to be honest I'd rather pick up smaller cases to get familiar with vue first. I think also kaj has experience with this, do you think you could pick it up @kajdehoop? If we don't do it all at once I could try to work on smaller parts of it perhaps; I'd also be happy to share some high-level ideas anyway!

kajdehoop:
Yes I can look into this issue ;) Maybe it should be split into two separate issues. We can use Vue i18n for the interface translations, but this is a separate issue from adding translation fields for the role information.

aleesaan:
Awesome! Indeed I could then pick up the other part, to update the role info with translation objects. For the multilingual input, I think we could use a field with an icon (maybe a flag?) on the side, to toggle the language or perhaps show another input below, with the other language. Just a thought, we can also discuss better later!

SamLubbers:
@kajdehoop Good idea to split this into two separate issues. I just made the other one here: #103 

@aleesaan good to know you'd rather work on smaller issues first. I've assigned you to issue #103, but this is too much to start with you can always look at other smaller features.

kajdehoop:
Note from the meeting: the cost of the translation service will be researched by @aleesaan

[...]

PR #124, opened 2020-12-12
Initial setup for i18n

aleesaan:
Added initial support for internationalization. I kept it simple not to increase the scope of the case, but these are the things need to follow up (in order):

1. Add the translation function (`$t()`) in every (user-facing) string in the app;
2. Setup a way to extract the translations (perhaps using [this package](https://www.npmjs.com/package/vue-i18n-extract-translations)?) on commits to master - the specifics are to be discussed;
3. Setup a way to push the translation files to a translation service (we're currently thinking of [Loco](https://localise.biz/)), as well as accounts for translators so to translate those files;
4. Setup a way to pull the (translated) translation files, possibly when building / deploying the app.

I will attach before/after screenshots once we finalize the design and fix the current issue with styles.

#39


aleesaan:
I just added some changes to complete the point 1 from above, and I also added the package mentioned in point 2 to extract them.

I still need to do two things in this case:
- Figure out how to properly handle missing translations, as currently strings with interpolations (e.g. `Published on {date}`) don't work;
- Figure out how to lazy load translations, as we shouldn't import the translation files in the main app bundle. I found [this guide](https://kazupon.github.io/vue-i18n/guide/lazy-loading.html) but couldn't get it to work, will need to look into it more.

kajdehoop:
@aleesaan Awesome, looks good! Thank you :D 

The issue also has point 3 & 4 about using a translation service like Loco to manage translations. Maybe we should merge this request first and then put point 3 & 4 into a new issue? @SamLubbers

kajdehoop (approved):

kajdehoop:
@SamLubbers I can't merge this PR as I don't have write access, can you do that? ;) Ale will create a new separate issue for the managing of the translations

SamLubbers:
Hey @kajdehoop good to see that you two have got the translations working!! I think that you do have write access since you're in the XRNL team for the volunteer platform. I think what is needed to merge this PR is for you to accept the changes (go to `file changed > review changes > approve > submit review`). If this doesn't work I'll see what I can change to give you write access :ok_hand:

PR #132, opened 2021-01-16
Conditional CI step

aleesaan:
Trying conditional CI steps.


aleesaan:
Merging this case to test that the [new workflow](https://github.com/xrnl/aginun/blob/380caf84cabb63d78e11feeff0beba1b541c7d48/.github/workflows/i18n.yml) runs only on commits to master.

PR #133, opened 2021-01-17
Fix CI

aleesaan:
Fixed CI after my previous commits, now we have a [job](https://github.com/xrnl/aginun/blob/b3e58cdbeb1487148333a24f78db696311da7187/.github/workflows/ci.yml#L69) that will only run on commits to master 🎊


SamLubbers:
@aleesaan I was going through my Github notifications and only just saw that you made this great change :raised_hands: It's so much better now without jobs duplicate among different files, and with a conditional logic for the `i18n-push` job.

Issue #122, opened 2020-12-05
Store auth token in front-end


kajdehoop:
## Problem that this feature solves
Right now if a user logs in, the authentication token is stored in the Vuex state. If the user closes the window or refreshes the page, this state is gone and the user needs to login again.

## Feature description

Describe a step-by-step use case of your feature. 

1. The token should be stored in local storage (cookie) after logging in
2. The token should be retrieved from local storage. If it exists the token should be set in the Vuex store, if it doesn't exist the user should not be logged in.

## Additional comments
Tokens have a lifetime, depending on settings in the KeyCloak keyserver. It would be best to have the token in the local storage expire after the lifetime of the token has passed, as the token won't work for requests anymore. Another option would be to invalidate the local storage if a request to the Hasura server fails because of an invalid/expired token, and thus have the user login again.

tomootes:
@aleesaan the issue im doing right (#130 ) is depending on this one. I found this tutorial which might be helpful: https://hasura.io/blog/best-practices-of-using-jwt-with-graphql/#basics_client_setup_expiration

Right now when i want to connect to Hasura using a token from Keycloak i get the error: `Could not verify JWT: JWTExpired`. Maybe we can have a quick call about this? It all seems quite connected.

Issue #121, opened 2020-12-05
Trigger login modal by using the login button in the app bar


kajdehoop:
## Problem that this feature solves
Makes it possible to login using existing username and password

## Feature description

Describe a step-by-step use case of your feature. 

1.  Clicking the login button on the app bar opens a modal
2. This modal has a username and password input field, and a login button
3. The login button dispatches the user/login action with the username and password as arguments

DenBaas:
I started working on this today and it is a good issue to start :) 

There is a problem I encountered that is a 503 http error I get when trying to log in: "CORS missing allow origin" and I believe this is a server side issue (because of [https://forum.vuejs.org/t/no-access-control-allow-origin-header/36995](this answer) for example). Can you help me with this?

SamLubbers:
@DenBaas awesome that you've started working on this issue and sorry you already ran into a problem that blocked you from making progress.

@kajdehoop could you check if there is an issue with the Authentication server that you set up? The login used to work when we merged the code so I assume it's also a problem on the server side.

Bas, until the issue is fixed, you can start creating the login front-end while faking the server response. To do that you can comment out the code inside the `login` action in the user Vuex store `src/store/modules/user.ts` and just call `commit("setToken", true);`

PR #134, opened 2021-01-25
Login modal & auth token

DenBaas:
This solves #121 and #122. The only thing left is styling the login form, because I did nothing about this.
afbeelding

What it does (more or less):

- Shows error message on failed login
- Can only send one login request simultaneously
- Saves login token in cookie (with expires field)
- login on enter
- resets fields after reopening the form


src/components/layout/LoginModal.vue
```diff
@@ -0,0 +1,89 @@
...
+</template>
+
+<script>
+  export default {
+    data () {
+      return {
+        name: "login-modal",
+        dialog: false,
+        username: "kaj-dev",
+        password: "test",
```
kajdehoop, line 49: These should be initialized as empty strings, instead of the 'kaj-dev' and 'test'

src/components/layout/LoginModal.vue
```diff
@@ -0,0 +1,89 @@
...
+        e.preventDefault();//prevents sending login request
+      },
+      login(e) {
+        var self = this;
+        e.preventDefault();
+        const { username, password } = this;
+        if(!this.readyToLogIn)
+          return;
+        this.waitingForServer = true;
+        this.$store.dispatch("user/login", {username, password, self}).then(result => {
```
kajdehoop, line 71: Why is `self` set to `this` and then send to the `user/login` action? It doesn't seem to be used there. 

Another small comment on best practices: it's best to use `let` or `const` instead of `var` for variables, as `var` is being phased out in newer JS versions. `let` is used for variables that contain a value that will later change, `const` is used for variables that contain a value that never changes. `var` can have some unexpected effects due to its scope. [Differences between var, let and const](https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/)

kajdehoop (changes requested):
@DenBaas Looks good Bas, thank you so much! 😄 

I've added a couple of comments on your code. In my local instance I had one problem, I didn't see the error messages. This was due to a CORS issue. I think this has nothing to do with your code but is due to a configuration issue on the keyserver. I'll check if we can fix that 🙂 

Great that the login can be completed by pressing the enter key! I have one other suggestion to improve the user experience: when opening the login modal, the username field should be in focus, so people don't have to click in that input field to start typing (they can then just type username -> press tab -> type password -> press enter).

I've attached the design for the login modal. Do you have some experience with CSS to finish the styling (or want to give it a try)? Or should I fix the styling for the modal?

image

src/components/layout/LoginModal.vue
DenBaas: > 
> 
> @DenBaas Looks good Bas, thank you so much! 😄
> 
> I've added a couple of comments on your code. In my local instance I had one problem, I didn't see the error messages. This was due to a CORS issue. I think this has nothing to do with your code but is due to a configuration issue on the keyserver. I'll check if we can fix that 🙂
> 
> Great that the login can be completed by pressing the enter key! I have one other suggestion to improve the user experience: when opening the login modal, the username field should be in focus, so people don't have to click in that input field to start typing (they can then just type username -> press tab -> type password -> press enter).
> 
> I've attached the design for the login modal. Do you have some experience with CSS to finish the styling (or want to give it a try)? Or should I fix the styling for the modal?
> 
> image

Thanks for the feedback! I can make this :)

DenBaas: I'm not sure why that response got there. buy anyway: we should not forget to remove this and change/remove this account for security. I'll put it back for now

src/components/layout/LoginModal.vue
DenBaas: alright. thanks again :)

I want the login button to be disabled while the user is waiting for the server to respond to the login request. this is why i made `waitingForServer`. to use that i needed `self` because it is set in `then(..)`. could i use `await` instead?

kajdehoop: I think `this.waitingForServer` can be used in the `then()`, as you've used an arrow function. You're also using `this.errorMessage` there so the `this.waitingForServer` should work in the same way. Correct me if I'm wrong though :) 

Btw another thing that I now see, you don't need to pass the event `e` to the methods and do preventDefault(). In vue you can do this by using the prevent modifier, for example `@click.prevent="login"`. Should work in the same way, but makes the code a bit cleaner and simpler 🙂 I also see that in the dialog watcher, you pass `d` into the function. I'm not 100% sure, but I think `d` is the value of the variable right? I would then just call it `value` or `val` to make it more clear.

kajdehoop (changes requested):

src/components/layout/LoginModal.vue
DenBaas: Do you have the specific measurements/sizes or do you know where I can find it?

kajdehoop: Good question! Here are the most important ones:
Title fontsize: 20px
Text fontsize: 14px
Width modal: 365px
Height modal: 285px
Width inputfields: 200px
Height inputfields: 30px
Login button fontsize: 20px
Button color: #3A62A8

A padding on the modal of around 20-25px should be fine. I think the login button can be made a bit smaller (incl the fontsize), maybe 16px fontsize.

I'll see if for the new designs I can find a way to share them through Adobe XD so you can see the sizes of all elements :)

DenBaas: afbeelding
thanks! I created this now and I think the text is small. What do you think? The input fields don't match yet with the inputs in the modal to create a new role:
afbeelding
i will still do that

src/components/layout/LoginModal.vue
DenBaas: thanks!

kajdehoop (approved):

kajdehoop (approved):

kajdehoop (approved):

kajdehoop:
Everything looks nice ^^ @SamLubbers I did the review, but see there are conflicts in package-lock.json and package.json. Could you review those two? Not completely sure what to do about them ;)

aleesaan:
I can pick it up from here Kaj! I also wanted to mention a few things 🤓

src/store/index.ts
```diff
@@ -31,5 +34,8 @@ const store = new Vuex.Store<RootState>({
   // https://vuex.vuejs.org/guide/strict.html
   strict: process.env.NODE_ENV !== "production"
 });
+if (Vue.$cookies.isKey("loginToken")) {
```
aleesaan, line 37: I would move this initialization inside the `created` method in the main.ts file, where we initialize the Vue instance.

src/store/modules/user.ts
```diff
@@ -1,27 +1,43 @@
...
+      return !!state.token;
+    }
   },
   mutations: {
-    setToken(state, token) {
+    setToken(state, token: string) {
       state.token = token;
+    },
+    setTokenCookie(state, { token, lifetime }) {
+      Vue.$cookies.set("loginToken", token, lifetime);
```
aleesaan, line 23: Instead of repeating this `"loginToken"` string (which can lead to errors), we create and export a const from the top of this file:

```ts
export const loginCookieKey = "loginToken";
```

Then we can use it in this file, as well as import it in other files as needed.

src/store/index.ts
```diff
@@ -1,13 +1,15 @@
...
 
 import alerts from "./modules/alerts";
 import roles from "./modules/roles";
 import groups from "./modules/groups";
 import errors from "./modules/errors";
 import user from "./modules/user";
+import VueCookies from "vue-cookies";
 
 Vue.use(Vuex);
+Vue.use(VueCookies);
```
aleesaan, line 12: This file is specific to the Vuex store, so I wouldn't add this here. We can instead move it to the main.ts file.

src/store/modules/user.ts
```diff
@@ -1,27 +1,43 @@
-import axios from "axios";
 import qs from "qs";
+import axios from "axios";
+import Vue from "vue";
+import VueCookies from "vue-cookies";
+
+Vue.use(VueCookies);
```
aleesaan, line 6: We don't need to call this here again, since we already call it elsewhere.

src/store/modules/user.ts
```diff
@@ -1,27 +1,43 @@
...
 
 export default {
   namespaced: true,
   state: {
     token: ""
   },
   getters: {
-    loggedIn: (state) => !!state.token
+    loggedIn(state) {
```
aleesaan, line 14: Can you revert this change? It's good to use arrow functions when possible, as they're provide a shorter syntax.

src/store/modules/user.ts
```diff
@@ -1,27 +1,43 @@
...
+      Vue.$cookies.set("loginToken", token, lifetime);
+    },
+    removeToken(state) {
+      state.token = null;
+      Vue.$cookies.remove("loginToken");
     }
   },
   actions: {
-    async login({ commit }, { username, password }) {
+    setTokenOnStart({ commit }): void {
```
aleesaan, line 31: What do you think of `initializeFromCookie` as the name? Also, you don't need to specify `: void`, as typescript already infers it automagically ✨

src/store/modules/user.ts
```diff
@@ -1,27 +1,43 @@
...
+      commit("setToken", Vue.$cookies.get("loginToken"));
+    },
+    async login(
+      { commit },
+      { username, password }
+    ): Promise<[boolean, string]> {
```
aleesaan, line 37: Same here, we don't need to specify the return type as it gets inferred! You can see this by removing it and then hovering on the function, you'll see that TS recognizes it correctly.

src/store/modules/user.ts
```diff
@@ -30,19 +46,27 @@ export default {
...
-      commit("setToken", token);
+      let success = false;
+      let message = "";
+      await axios
+        .post(
+          process.env.VUE_APP_KEYSERVER_URL || "",
+          qs.stringify(params),
+          config
+        )
+        .then(function(res) {
```
aleesaan, line 57: Since you use `await` above, you don't need to use `then` here. This can become:

```ts

try {
  const { data } = await axios.post(
    process.env.VUE_APP_KEYSERVER_URL || "",
    qs.stringify(params),
    config
  );
  commit("setToken", data.access_token);
  commit("setTokenCookie", {
    token: data.access_token,
    lifetime: data.expires_in
  });
} catch ({ response }) {
  return response.data?.error_description || "Login server unavailable";
}
```
I dropped the `success` variable because it wasn't used outside this function. If everything goes smooth, we simply commit stuff, whereas if there's an error, we return it so it can  be shown to the user.

src/styles/global.scss
```diff
@@ -17,9 +17,84 @@
...
+    input {
+      width: 200px;
+      height: 30px;
+      //padding: 0 20px 0 20px;
+      text-align: left;
+      align-self: center;
+    }
+    #submit-button {
+      font-size: 20px;
+      background-color: #3a62a8;
```
aleesaan, line 93: We don't need this, the button already uses the primary color.

src/styles/global.scss
```diff
@@ -17,9 +17,84 @@
...
+    display: flex;
+    justify-content: center;
+    form {
+      padding: 0;
+      margin: 0;
+    }
+    a {
+      align-self: center;
+    }
+    input {
```
aleesaan, line 84: I think we can remove this declaration altogether, I don't see any difference without it.

src/styles/global.scss
```diff
@@ -17,9 +17,84 @@
...
+  font-family: "FucXed";
+  padding: 0.5rem 1rem;
   box-shadow: none;
 }
 
+.error-message {
+  display: inline-block; 
+  align-self:center; 
+  margin-bottom: 15px;
+  span {
```
aleesaan, line 30: Can you leave an empty line before each new declaration?

src/styles/global.scss
```diff
@@ -17,9 +17,84 @@
...
+    color: #C20000;
+    background-color: #FFBCBC;
+    border-radius: 5px;
+    text-align: center;
```
aleesaan, line 34: I think we can remove this and the `align-self` below, I don't see difference without them.

src/styles/global.scss
```diff
@@ -17,9 +17,84 @@
...
-  font-family: 'FucXed';
-  padding: .5rem 1rem;
+.v-btn {
+  font-family: "FucXed";
+  padding: 0.5rem 1rem;
   box-shadow: none;
 }
 
+.error-message {
+  display: inline-block;
```
aleesaan, line 27: I think we can remove this, I don't see difference without it.

src/styles/global.scss
```diff
@@ -17,9 +17,84 @@
...
+    border-radius: 5px;
+    text-align: center;
+    align-self: center;
+    padding: 4px 10px 4px 10px;
+  }
+}
+
+#login-modal {
+  font-size: 14px;
+  //padding: 0 20px 20px 20px;
```
aleesaan, line 42: Can you remove this?

src/styles/global.scss
```diff
@@ -17,9 +17,84 @@
...
+  padding: 20px;
+  v-input {
```
aleesaan, line 44: I don't see this declaration being used, can we drop it? Unless I'm missing something?

src/styles/global.scss
```diff
@@ -17,9 +17,84 @@
...
+    margin: 0;
+    padding: 0;
+  }
+  p {
+    margin-bottom: 10px;
```
aleesaan, line 49: We have another `p` declaration below, can you move this there?

src/styles/global.scss
```diff
@@ -17,9 +17,84 @@
...
+  }
+  #close-button {
+    position: absolute;
+    right: 6px;
+    font-size: 10px;
+    top: 15px;
+    margin-right: 10px;
+  }
+  #close-button::after {
+    content: " \2573";//2715
```
aleesaan, line 63: Can you remove the comment here?

src/styles/global.scss
```diff
@@ -17,9 +17,84 @@
...
+  }
+  p {
+    width: 240px;
+    text-align: center;
+    align-self: center;
+  }
+  .container {
+    margin: 0;
```
aleesaan, line 71: I think we can drop `margin`, `flex-wrap`, and `justify-content` from here, I don't see any difference without them.

src/styles/global.scss
```diff
@@ -17,9 +17,84 @@
...
+    padding: 0;
+    flex-wrap: wrap;
+    flex-direction: column;
+    display: flex;
+    justify-content: center;
+    form {
```
aleesaan, line 77: We can drop these styles, they're not being used since the `form` isn't inside the `.container` element.

aleesaan (changes requested):
Hey @DenBaas thank you so much for this PR!  
I'm leaving quite a lot of comments but don't get overwhelmed by the number, most of them are very simple changes (like removing some extra code), while in the rest of the comments I tried to leave sample codes (and link some relevant resources) to make it easier for you to implement them.  
Feel free to answer my comments and let me know if something isn't clear!

src/components/TheAppBar.vue
```diff
@@ -57,13 +55,16 @@ import { mapGetters, mapActions } from "vuex";
 import { contactEmail } from "@/constants/contacts";
 import styles from "@/constants/styles";
 import LanguageSelect from "@/components/LanguageSelect.vue";
+import LoginModal from "@/components/layout/LoginModal.vue";
+import store from "../store";
```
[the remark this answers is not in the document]
DenBaas: loginmodal is used, store not. fixed it

src/styles/global.scss
DenBaas: for some reason i need it

src/components/layout/LoginModal.vue
```diff
@@ -0,0 +1,99 @@
...
+  methods: {
+    cancel() {
+      this.dialog = false;
+    },
+    login() {
+      const { username, password } = this;
+      if (!this.readyToLogIn) return;
+      this.waitingForServer = true;
+      this.$store
+        .dispatch("user/login", { username, password, self })
```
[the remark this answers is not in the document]
DenBaas: i removed it

src/store/modules/user.ts
DenBaas: thanks! this works a lot better :)

src/styles/global.scss
DenBaas: i need this here as well for some reason

src/components/TheAppBar.vue
```diff
@@ -13,9 +13,7 @@
       <h2>{{ $t("Vacatures") }}</h2>
     </v-toolbar-title>
     <v-spacer />
-    <v-btn v-if="!loggedIn" text @click.stop="login">
-      {{ $t("Login") }}
-    </v-btn>
+    <login-modal ref="loginModal" v-if="!loggedIn"></login-modal>
     <v-btn v-else text @click.stop="logout">
```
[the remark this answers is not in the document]
DenBaas: I agree that the logic is mixed now and should be better. The logout button is here while the login button is in `LoginModal.vue`. Why should I not put it all in the login modal?

.gitignore
```diff
@@ -27,3 +27,4 @@ yarn-error.log*
 apollo.config.js
 
 *.wiki
+.eslintrc.json
```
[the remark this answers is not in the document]
DenBaas: That file was created while setting up vscode so i added it since it is specific for me?

src/components/TheAppBar.vue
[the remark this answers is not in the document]
aleesaan: Indeed, that's also a possibility! But then it wouldn't be only a `LoginModal`, as it would contain also buttons and logic to open/close it, as well as logic to know whether you're authenticated or not (to show the right button). That's why I suggested the above, I think it provides a better separation of concerns - the `LoginModal` component will only contain/show the modal (and its contained logic), and the rest can be handled outside. What do you think?

.gitignore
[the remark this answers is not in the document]
aleesaan: I think you can safely remove both the file and this line, cause we use `.eslintrc.js` in this project (and we do need to track it, as it provides the linting rules for the project).

src/components/TheAppBar.vue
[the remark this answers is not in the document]
DenBaas: Well I think it is better to keep `loginModalOpen` in the login modal, so what `dialog` is now. I can rename that. I agree it is not nice that the login and logout button are now mixed so to solve it I want to put it in `LoginModal` because for me that is the easiest for me now. I like to keep everything together as much as possible so that means the login/logout functions should be together.

package.json
```diff
@@ -20,55 +20,56 @@
     "apollo-client": "^2.6.10",
     "apollo-link": "^1.2.14",
     "apollo-link-http": "^1.5.17",
-    "axios": "^0.21.0",
-    "core-js": "^3.7.0",
+    "axios": "^0.21.1",
```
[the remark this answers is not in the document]
aleesaan: There are still unneeded changes in the package.json (and package-lock.json), can you please have a look? Only keep the change for `vue-cookies`.

src/components/layout/AuthState.vue
```diff
@@ -0,0 +1,113 @@
...
+      this.errorMessage = "";
+    },
+    close() {
+      this.dialog = false;
+    },
+    ...mapActions("user", ["logout"]),
+    async login() {
+      const { username, password } = this;
+      this.serverLoading = true;
+      const message = await this.$store.dispatch("user/login", {
```
aleesaan, line 100: Can you call this `errorMessage` to be more explicit, and below directly check for `if (!errorMessage) { ... }`?

src/components/layout/AuthState.vue
```diff
@@ -0,0 +1,113 @@
...
+      );
+    },
+    ...mapGetters({
+      loggedIn: "user/loggedIn"
+    })
+  },
+  methods: {
+    open() {
+      this.dialog = true;
+      this.username = ""; //kaj-dev
```
aleesaan, line 88: I think it would be better if we reset these properties after a successful login, rather than on `open()` - so that if you mistakenly close the modal, and reopen it, they're still there. I'd suggest the following:

- Remove the `open` method altogether, and directly do `@click="dialog = true"` in the template
- Create a `reset` method that resets username, password, serverLoading and errorMessage, and call it after a successful login
- Bonus, I'd remove the `close` method too, since it's only used once, and instead directly do:

```ts
if (!errorMessage) {
  this.dialog = false;
  this.reset();

  return;
}
```

src/store/modules/user.ts
```diff
@@ -1,48 +1,71 @@
...
+          qs.stringify(params),
+          config
+        );
+        commit("setToken", data.access_token);
+        commit("setTokenCookie", {
+          token: data.access_token,
+          lifetime: data.expires_in
+        });
+      } catch ({ e }) {
+        if (e) {
```
aleesaan, line 58: This will never be true, cause `e` isn't a property that you can destructure from the error. Can you please have a look at my [previous comment](https://github.com/xrnl/aginun/pull/134#discussion_r567295654) about how we should do this?

aleesaan (changes requested):

src/store/modules/user.ts
DenBaas: idk why i changed it back. sorry!

src/components/layout/AuthState.vue
```diff
@@ -0,0 +1,113 @@
+<template>
+  <v-dialog
+    v-model="dialog"
+    width="365px"
+    height="285px"
+    @click:outside="$emit('close')"
```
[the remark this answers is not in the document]
DenBaas: which one?

src/store/modules/user.ts
```diff
@@ -1,48 +1,71 @@
...
     async login({ commit }, { username, password }) {
       const config = {
         headers: {
-          "Content-Type": "application/x-www-form-urlencoded"
+          "Content-Type": "application/x-www-form-urlencoded",
         }
       };
-
       const params = {
-        // eslint-disable-next-line @typescript-eslint/camelcase
```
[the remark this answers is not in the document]
DenBaas: i get this error here if i add it: `Definition for rule '@typescript-eslint/camelcase' was not found  @typescript-eslint/camelcase`

src/components/TheAppBar.vue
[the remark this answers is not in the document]
SamLubbers: @DenBaas hey I see that you've made a lot of progress with this pull request :raised_hands: I'll also share some of my feedback in the areas where I think it can help or where there is no agreement.

While I see that at first it can be easier to write all the related functionality in one place, when this is done continuously it leads to bloated functions/components that are doing too much functionality, making them more difficult to understand and reuse. To avoid these issues it's good to stick to important programming design principles such as the principle of [single responsibility](https://en.wikipedia.org/wiki/Single-responsibility_principle) & [modularity](https://en.wikipedia.org/wiki/Modular_programming). 

The suggestion from @aleesaan supports these principles:

* single responsibility: the modal component only has the logic to let a user input their credentials and login, and nothing else. With the code that he suggested, someone that reads `TheAppBar` will immediately see that there is one button, one modal and that the button controls the modal. When grouping all of that together it's not immediately obvious.
* modularity: since the component only has a modal, and not the buttons, it's more modular and more easily reusable and interchangeable. For example, if you want to activate the modal from a different place in mobile devices, you can easily do so, by adding a separate button on small screens (just like we do for the creation of new roles). Or if we like the button but want to replace the modal, we can just keep the buttons in the code and replace the modal with a different component.

I hope this clarifies Ale's suggestion a bit. Please share if you still have questions :ok_person:

src/styles/global.scss
SamLubbers: hmm that's curious :sweat_smile: @DenBaas can you share what happens with/without this code?

src/styles/global.scss
SamLubbers: @DenBaas here I'll share the same feedback as the other comment above. If you share what happens in your computer then we can try to reproduce it and understand why it is needed :ok_hand:

src/store/modules/user.ts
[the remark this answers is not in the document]
SamLubbers: @DenBaas hmmm that's interesting. I think the issue might be that you have a newer version of `typescript-eslint`. [Here](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/camelcase.md) I see that the camelcase rule has been deprecated in favour of the naming-convention rule. Can you try using this rule instead?

src/components/layout/AuthState.vue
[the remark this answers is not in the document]
SamLubbers: @DenBaas I'll unresolve this conversation as I think there is still no conclusion on this one. If you implement the feedback you can mark it as resolved, but if you don't it's useful if you share why you think the suggestion doesn't need to be implemented. That way each person reviewing the code has a good understanding of why something has/hasn't be incorporated into the code :nerd_face:

src/store/modules/user.ts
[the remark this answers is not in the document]
SamLubbers: Actually I see now in the `package.json` that you've upgraded the `@typescript-eslint/eslint-plugin` package. That's probably leading to the error. Like Ale say's [here](https://github.com/xrnl/aginun/pull/134#discussion_r567262955) it is best if you don't upgrade the packages in this pull request and only add the one new package you've added.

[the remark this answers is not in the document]
aleesaan: Good catch, I think this might fix the issue indeed!

src/components/layout/AuthState.vue
```diff
@@ -0,0 +1,113 @@
...
+      username: "",
+      password: "",
+      errorMessage: "",
+      serverLoading: false
+    };
+  },
+  computed: {
+    buttonDisabled: function() {
+      return (
+        this.username.trim().length == 0 ||
```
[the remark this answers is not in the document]
SamLubbers: Actually, what will work even better would be to use [vee-validate](https://vee-validate.logaretm.com/v3/guide/basics.html) in order to perform the validation of the input fields. I suggest you read through the basics of the documentation and then implement the validation for the username and password field. Not only will it check for many more conditions with fewer code, but it will also makes it easy to determine the [validation state](https://vee-validate.logaretm.com/v3/guide/state.html) so that you can disable a button when the fields are not valid. 

In the [Vuetify documentation](https://vuetifyjs.com/en/components/forms/#vee-validate) you can also see good examples on how you can use vee-validate with Vuetify to provide nice error messages with Vuetify components.

src/components/layout/AuthState.vue
```diff
@@ -0,0 +1,113 @@
...
+import { mapGetters, mapActions } from "vuex";
+export default {
+  name: "AuthState",
+  data() {
+    return {
+      dialog: false,
+      username: "",
+      password: "",
+      errorMessage: "",
+      serverLoading: false
```
SamLubbers, line 70: I see that you use a `serverLoading` prop to disable the button on submit. This is already very useful to show the user that they cannot log in again :ok_hand: To make it even better I would include a spinner to show the user that the application is loading and that they need to wait. We already have the [saeris/vue-spinners](https://github.com/saeris/vue-spinners) package installed, which makes it really easy to add nice looking spinners

src/components/layout/AuthState.vue
```diff
@@ -0,0 +1,113 @@
+<template>
+  <v-dialog
+    v-model="dialog"
+    width="365px"
+    height="285px"
+    @click:outside="$emit('close')"
+  >
+    <template v-slot:activator="{}">
```
SamLubbers, line 8: Using the v-slot:activator is only necessary when you use the `{ on }` functionality to trigger opening the dialog

```vue
      <template v-slot:activator="{ on, attrs }">
        <v-btn
          color="red lighten-2"
          dark
          v-bind="attrs"
          v-on="on"
        >
          Click Me
        </v-btn>
      </template>
```
When you use a separate function you can just keep the button outside. I personally think this is clearer, because at the top level you will have a button to trigger the modal, and then the modal itself.

SamLubbers (changes requested):
Hey @DenBaas, thank you for implementing all of this functionality :raised_hands: I can imagine it was a looooot of work, especially considering you joined the project not too long ago!

I first reviewed the existing comments in the code and added my feedback where it was relevant. Then I went through the entire code to add some additional feedback, so that you can also get some of my insights and not just Ale's :stuck_out_tongue_winking_eye: 

Besides the feedback that I've left in the code, the last thing that I'd like to add is that it would be great to have tests for the new `AuthState` component and for the updated functionality in `store/user.ts`. This might be a bit too much for in this pull request, so it might be better if you can make a new issue so that we make sure we write tests for it in the future :nerd_face:

package.json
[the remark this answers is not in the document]
DenBaas: i checked out `package.json` and `package-lock.json` to the version on the master and installed `vue-cookies` and the build works now. is it correct now?

src/store/modules/user.ts
DenBaas: i reverted it now!

src/styles/global.scss
DenBaas: well i'm implementing the vuetify components now and this css can be removed. i will show the differences next time

src/styles/global.scss
DenBaas: thanks. same as in the other comment :)

src/components/layout/AuthState.vue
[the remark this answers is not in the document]
DenBaas: thanks! i think i got it correct now. a new issue now is that error messages are weirdly translated:
afbeelding
what should i do with this?

src/components/layout/AuthState.vue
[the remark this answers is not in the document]
DenBaas: we talked about it in private. it turned out that `@click:outside` was not necessary because the modal closes when clicking outside of the modal without this line.

package.json
```diff
@@ -74,5 +76,20 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/DenBaas/aginun.git"
+  },
+  "author": "",
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/DenBaas/aginun/issues"
+  },
+  "homepage": "https://github.com/DenBaas/aginun#readme"
```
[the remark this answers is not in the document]
DenBaas: it is deleted now. idk how it got here haha

src/styles/global.scss
```diff
@@ -17,9 +17,81 @@
...
+    color: #c20000;
+    background-color: #ffbcbc;
+    border-radius: 5px;
+    text-align: center;
+    align-self: center;
+    padding: 4px 10px 4px 10px;
+  }
+}
+
+#login-modal {
```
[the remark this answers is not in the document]
DenBaas: i still have the css for the close button because i think it is the most clean solution. i tried adding it with rows but then the title is not centered anymore. do you have a solution or can i just keep it like that?

aleesaan:
Closing this PR because of inactivity, I will open a new one to finalize this case.

Issue #26, opened 2020-02-16
Filters in URL


SamLubbers:
When a user selects/deselects filters, the URL should be automatically updated with the filter information. Then, when that URL is shared with others, the page should immediately load the tasks/roles with the relevant filters.

This will enable local groups and working groups to automatically link to available positions within their groups. For example, the Outreach & Training group in Amsterdam might have a link in their Mattermost channel header which new people can click on to automatically view available positions.

PR #135, opened 2021-01-26
URL structure & filters

DenBaas:
I think everything should work now and I hope the routes are fine too.

The parameters are parsed from a JSON to a string which is stored in the url. Perhaps it needs some renaming etc. but otherwise it is done.

- [x] updating parameters in the url
- [x] searching without parameters clears the url
- [x] the parameters from the url are put into the filters on load
- [x] empty parameters (like an empty role title or no groups selected) are not put in the url
- [x] time commitment is only stored if it is different from [min, max]

todo:
- [x] keep filters in url (and not delete them) when clicking on role and going back
- [x] update url after "clear all filters"


SamLubbers:
@DenBaas thank you for making this great change :smile: :raised_hands: These filters in the URL will make it so much easier to share and find roles.

Since I see that you're still working on #134 I'll first share my feedback there. Once that pull request is completed and merged I'll do the review of this pull request so that you can continue making progress :ok_person:

SamLubbers:
Closing this PR because it has diverged too much from the master branch. If someone else wants to implement #26 they can use this branch as a basis, and apply the main changes on top of the head of the master branch.

PR #141, opened 2021-02-15
Multlingual input

### LINKED ISSUE #103: Multilingual role publishing, editing and viewing

Roles should be published in both Dutch and English, so that all rebels in XR NL can read about them and apply. Currently it's only possible to post the roles in one language.

### Role creation and editing

When a rebel publishes a new role they should provide these fields in both Dutch and English:
* Title
* Responsibilities
* Description
* Requirements

All other fields can be automatically translated through string translation. To start we'll just keep it simple and add all fields to one form, so we'll have first description (Dutch) and then description (English), and so for all the fields. This format also makes it easy to edit the roles in different languages.

On submit, the dutch and english text is then stored in the roles table of the database.

Rebels should also be able to mark a role as only relevant for dutch or english speakers, which would then make the fields of the other language not-required.

In the future we can improve this by only requiring information in one language, and when users click submit, they are shown a set of fields to be translated. What would make this even easier for users if we already automatically translated everything using DeepL, and they just need to adjust the translated text.

#### Roles overview

When user are on the Dutch version of the page, they see all roles with the dutch text. Same for the english version of the page. If a role does only have a dutch version, it shouldn't appear on the english version of the page, and vice versa.

### Individual role view

A last edge case to consider is if a user is in the english version of the page, and go to a URL corresponding to a role that is only in Dutch. In such cases the role view should display a text saying that this role is only available in dutch, along with a button to go to the dutch version of the page. For this you can reuse the functionality for when a role is already taken.

aleesaan:
Created a multi-language input to enable multilingual roles.

Resolves #103.


src/components/MultiLanguageInput.vue
```diff
@@ -0,0 +1,122 @@
...
+        v-else
+        outlined
+        dense
+        class="rounded-lg"
+        :value="internalValue.nl"
+        @input="updateValue('nl', $event)"
+        :error-messages="errors"
+        :placeholder="placeholder"
+      >
+        <template v-slot:label>🇬🇧 {{ $t("Dutch") }}</template>
```
SamLubbers, line 63:
```suggestion
        <template v-slot:label>🇳🇱 {{ $t("Dutch") }}</template>
```

src/components/roles/RoleEditDialog.vue
```diff
@@ -3,189 +3,187 @@
...
-                  <v-icon>mdi-close</v-icon>
-                </v-btn>
-              </v-list-item>
-            </template>
-          </v-card>
-          <validation-provider
-            v-slot="{ errors }"
+          <multi-language-input
+            v-model="role.title"
+            rules="required"
```
SamLubbers, line 25: I would add the rules it had before. This prevents people from entering extremely long titles that don't fit in the role overview card. Also it's good to prevent special characters from appearing in the title.
```suggestion
            rules="required|alpha_spaces|max:30"
```

src/components/roles/RoleEditDialog.vue
```diff
@@ -219,45 +217,16 @@ import {
...
-    } else {
-      article = "a";
-    }
-
-    return i18n.t(`You must specify {article} {fieldName}.`, {
-      article,
-      fieldName
-    });
+  message: () => {
+    return i18n.t("This field is required.");
```
SamLubbers, line 227: Wow what a massive clean up of the validation :clap: :clap: When I wrote the previous code I think I was a bit too excited about passing the field names to the error messages :laughing:

SamLubbers (changes requested):
@aleesaan what an insanely big improvement on the translations :star_struck: It doesn't only work almost perfectly, but the code is also so clean and well organised. 

These are the last things that I've seen that we need to change:

- [x] unable to press enter in the text areas. This would be useful so that
  people can write lists. 
- [ ] when a role is published, and the user clicks on new role, the
multilingual data from the previous role is still there. The current
`resetState` doesn't seem to clear the data for the translatable fields. 
- [x] show dutch as the first language, since the majority of people in the
movement have a preference for dutch. 
- [ ] imagine someone starts filling in a role in both languages and then
decides that they only want it in dutch. In that case, even if they de-select
the language, the data of both languages is still sent, and therefore still
appears when searching for roles. What is the best way to fix it? 1) delete the data of the unselected language before sending it to the backend? 2 have a field for each role specifying the languages for which the role should be displayed? I personally think that option 1 is better because option 2 will increase the complexity of the code to display roles.

A few last things that I need to fix in the backend:

- [x] send the selected language to search_roles (through a header)
- [x] do not show roles in a certain language if there is no data for that language
- [x] do not display an individual role view for roles that don't exist in that
  language. This might happen if people find a dutch only role in the dutch
  version of the page, share the URL, and someone else tries to see the role
  from their english version of the page.
- [ ] set multilingual data in backend for working circles, and display the correct language in all parts of the application.

src/components/roles/RoleEditDialog.vue
aleesaan: 😂

package-lock.json
```diff
@@ -16552,6 +16517,87 @@
         }
       }
     },
+    "vue-loader-v16": {
```
SamLubbers, line 16520: why did this file get updated? :thinking:

src/components/MultiLanguageInput.vue
```diff
@@ -119,8 +121,10 @@ export default Vue.extend({
     }
   },
   watch: {
-    value(value) {
+    value(value: Translation) {
       this.internalValue = value;
+      (this.$refs.nl_form as InstanceType<typeof ValidationProvider>).reset();
+      (this.$refs.en_form as InstanceType<typeof ValidationProvider>).reset();
```
SamLubbers, line 128: This code throws the following error if you publish a language with only one selected role. 

```
Error in callback for watcher "value": "TypeError: this.$refs.nl_form is undefined"
```

I think a simple if statement could fix it

src/components/roles/RoleEditDialog.vue
```diff
@@ -353,7 +361,17 @@ export default {
         this.editRole ? this.$t("Role edited") : this.$t("Role created")
       );
     },
-    isEmpty: (text) => !text || text.length === 0 || !text.trim()
+    isEmpty: (text) => !text || text.length === 0 || !text.trim(),
+    parseTranslation(translation) {
+      return (
+        translation &&
```
SamLubbers, line 368: out of curiosity, why is this line needed? Isn't it possible to directly return the result of ` Object.entries(translation)...`?

src/components/MultiLanguageInput.vue
```diff
@@ -117,6 +117,11 @@ export default Vue.extend({
       this.internalValue[key] = value;
```
SamLubbers, line 120: I think we can do this without having an `interalValue` variable:

1. clone `this.value` into `const newValue`
2. set `newValue[key] = value;`
3. `this.$emit("input", newValue);`

This way you can remove `this.internalValue = value;` within the watch statement, and directly read the data from `value`. `:value="value.en"`

SamLubbers (changes requested):
@aleesaan thanks so much for the last fixes. We're almost there!! :grin: There are a few last things to fix:

- [ ]  Currently when you enter multiple lines in the role description,
  responsibilities or requirements, they appear as one line in the
  `RoleViewDialog`. They should be rendered as multiple lines
- [ ] Currently, when you select no application deadline (`dueDate`), it is set
  to the current date. It should instead publish `dueDate` as `null` if no
  dueDate is selected. 
- [ ] When you edit a role the first time the right data appears in the form.
  But when you edit a role, click publish and then try to edit it again you
  won't see the role details anymore, instead it appears as a new role. 

I've also left some questions/suggestions in the code

package-lock.json
aleesaan: Not sure, probably related to the other dependency you added - I now deleted the node_modules + package-lock and did a fresh `npm i`; there's still a bunch of changes in the package-lock but I wouldn't worry as long as everything works fine.

SamLubbers (dismissed):
@aleesaan with your and my last couple of commits all the issues (that we know of) are now resolved :partying_face: 

Now the edit role form always works well and the right working circle colours are loaded from the backend. If you want you can test it one more time and if there are no more big issues you can merge it :raised_hands:

SamLubbers (approved):

Issue #147, opened 2021-03-28
Error handling for authentication


SamLubbers:
### case 1: missing or wrong VUE_APP_KEYSERVER_URL in `.env`

1. remove or change `VUE_APP_KEYSERVER_URL` from `.env`
2. click on `login`

### case 2: wrong credentials

1. change the credentials in `src/store/modules/user.ts` with invalid login details
2. click on login

## Expected result after following these steps

Display an alert with a useful error message (server is inaccessible, wrong credentials are used...)

## Actual result

No feedback. The user needs to open the console, which doesn't give a clear indication of what the problem is. 

## Additional comments

For all the alerts, you can use the existing `FeedbackAlert` component.

PR #150, opened 2021-04-17
Login modal and token in cookie

### LINKED ISSUE #121: Trigger login modal by using the login button in the app bar

aleesaan:
Resolves #121, #122 and #147.

## Additional comments
Follow-up for the [PR 134](https://github.com/xrnl/aginun/pull/134), which was closed due to inactivity.

### DIFF
--- a/src/components/AuthState.vue
+++ b/src/components/AuthState.vue
@@ -1,0 +1,133 @@
+<template>
+  <div>
+    <v-btn v-if="!loggedIn" @click="dialog = true">{{ $t("Login") }}</v-btn>
+    <v-btn v-else @click="logout">
+      {{ $t("Logout") }}
+    </v-btn>
+    <v-dialog v-model="dialog" :persistent="serverLoading" max-width="380px">
+      <v-card class="pa-3">
+        <div class="text-center">
+          <h2>{{ $t("Member login") }}</h2>
+          <p>
+            {{
+              $t(
+                "Log in to post new vacancies or edit existing ones. Don't have an account yet?"
+              )
+            }}
+            <a :href="`mailto:${contactEmail}`">{{ $t("Contact us") }}</a>
+          </p>
+          <validation-observer v-slot="{ invalid }" ref="login_form">
+            <form @submit.prevent="handleSubmit">
+              <validation-provider
+                v-slot="{ errors }"
+                name="username"
+                rules="required"
+              >
+                <v-text-field
+                  outlined
+                  dense
+                  class="rounded-lg"
+                  v-model="username"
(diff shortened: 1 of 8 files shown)


src/components/AuthState.vue
```diff
@@ -0,0 +1,133 @@
...
+                rules="required"
+              >
+                <v-text-field
+                  outlined
+                  dense
+                  class="rounded-lg"
+                  type="password"
+                  v-model="password"
+                  :label="$t('Password')"
+                  :error-messages="errors.length ? errors : serverError"
```
SamLubbers, line 48: I see that the `serverError` is now displayed in the password field. The issue that this creates is that it could make the user think that the problem is with the password, when in fact the problem could be the username. A better solution is to display some red text with the `serverError` either above or below the fields, or using a `FeedbackAlert`

src/store/modules/user.ts
```diff
@@ -31,18 +44,26 @@ export default {
...
+        commit("setTokenCookie", {
+          token: data.access_token,
+          lifetime: data.expires_in
+        });
+      } catch ({ response }) {
+        return (
+          (response.data?.error_description as string) ||
+          i18n.t("Login server unavailable")
+        );
+      }
```
SamLubbers, line 63: I thought that this would handle case 1 of issue #147, but when I change the endpoint of the auth server this error doesn't appear and instead an error appears in the console. If you can get this error handling to work then we could also close #147 with this PR

SamLubbers (changes requested):
@aleesaan thanks for implementing these two pull requests :raised_hands: :100: And thanks again for all the work you put into improving the previous pull request that fixed these issues.

The functionality works exactly like it should :ok_hand: I log in and out without a problem and I remain logged in even after refreshing the page.

I just left some comments in the code about small things that can be improved. On top of that, I think it would be good to display the `FeedbackAlert` when a user successfully logs in and out. Especially when logging out it happens really fast and some feedback would be useful to let the user know that it worked.

src/store/modules/user.ts
aleesaan: Done ✌️

SamLubbers (approved):
Everything works flawlessly :sparkler: Thanks @aleesaan!

I've just made the issue #154 so that we don't forget to clear the default user when deploying on production.

PR #152, opened 2021-04-23
Feature/add links navbar 97

### LINKED ISSUE #97: Add links to navbar
Right now we only have the (most important) functionality, but it would be nice if people can see some background info or FAQ about the platfom if they want to know more.

## Feature description
We can add a footer with a link to an about page and FAQ page. We can put the feedback button in there, to resolve issue #47.  A footer can also help with a bit of branding, to make the app look more professional and finished.

tomootes:

- New vanilla Navbar added
- New links added to the menu
- Added support and about page
- Navbar is now mobile friendly

- Vuetify navbar was hard to customize according design
- New pages were requested

Resolves #97 

## Additional comments

I added some new SCSS variables as well in `/styles/variables.scss`. Working like this, in my opinion, can create a better design system. I also added some "SCSS components" which are reusable parts for templating. These are situated in `/styles/components`. Let me know what you think about this.


package.json
```diff
@@ -15,6 +15,7 @@
     "test:unit:dev": "vue-cli-service test:unit --watchAll",
     "test:e2e": "vue-cli-service test:e2e",
     "lint": "vue-cli-service lint --no-fix",
+    "lint-fix": "vue-cli-service lint --fix",
```
SamLubbers, line 18: Nice that you added this command :ok_hand: Do you also have your editor configured to auto-fix eslint issues on save? I have that enabled and it makes coding much faster, as it automatically fixes prettier issues or other small issues (such as a missing `;`)

src/i18n/messages/nl.json
```diff
@@ -59,7 +59,11 @@
   "Login": "Inloggen",
   "Logout": "Uitloggen",
   "Support": "Hulp",
+  "Dashboard": "Overzicht",
```
SamLubbers, line 62: Did you create these translations manually? Or using the i18n commands listed in `package.json`? :thinking: 

If you've created them manually then I think we need better documentation explaining how we manage translations.

src/router/index.ts
```diff
@@ -23,6 +25,16 @@ const routes = [
...
+  {
+    path: "/about",
+    name: "about",
+    component: AboutPage
+  },
+  {
+    path: "/support",
+    name: "support",
+    component: SupportPage
+  },
```
SamLubbers, line 37: since all pages are already in one folder `views`, I think we can remove the `Page` from the component names. So these new components would become `About` and `Support`

src/styles/base/_typography.scss
```diff
@@ -13,4 +13,17 @@
 h1,h2,h3,h4,h5{
   font-family: 'FucXed';
   letter-spacing: 1px;
+}
+
+@mixin font-heading(){
```
SamLubbers, line 18: Is the `font-heading` intended to be used just in the navbar or in all kinds of headings? I think it would be good to clarify that in here, either with a comment or with a more explicit name for the mixing. For example: `font-navbar`

src/styles/global.scss
```diff
@@ -2,6 +2,12 @@
 @import "base/typography";
 @import "variables";
 
+// components
+@import "components/nav-link";
+@import "components/nav-bar";
+@import "components/container";
+@import "components/row";
```
SamLubbers, line 9: I see that these two reusable classes look a lot like the ones provided by [Vuetify's grid system](https://dev.vuetifyjs.com/en/components/grids). Would it be possible to use the Vuetify classes? Or is there an advantage to these new classes?

src/styles/global.scss
```diff
@@ -23,3 +29,6 @@
   box-shadow: none;
 }
 
+.v-content{
+  margin-top: 66px;
+}
```
SamLubbers, line 34: Is there a way for the navbar height to share this value `66px` with this class? That way, if we change the height of the navbar the margin will also be adjusted accordingly

src/styles/variables.scss
```diff
@@ -32,6 +32,20 @@ $border-radius-root: 0;
 $body-font-family: 'Crimson Text', serif;
 $font-size-root: 18px;
 
+$font-size-sm: 0.875rem;
```
SamLubbers, line 35: Are there equivalent helper functions from Vuetify that we could use for the variables you've defined here?

src/views/AboutPage.vue
```diff
@@ -0,0 +1,29 @@
...
+  </div>
+</template>
+<script>
+import { contactEmail } from "@/constants/contacts";
+export default {
+  name: "ErrorPage",
+  data: () => ({
+    contactEmail,
+    isDevMode: process.env.NODE_ENV !== "production"
+  })
```
SamLubbers, line 27: `name` must match the file name and data has to be removed as it's not used.

src/views/SupportPage.vue
```diff
@@ -0,0 +1,40 @@
...
+
+export default {
+  name: "ErrorPage",
+  components: {
+    IconLink
+  },
+  data: () => ({
+    contactEmail,
+    isDevMode: process.env.NODE_ENV !== "production"
+  })
```
SamLubbers, line 38: `name` must match the file name and `isDevMode` has to be removed as it's not used.

src/views/SupportPage.vue
```diff
@@ -0,0 +1,40 @@
...
+          :href="`mailto:${contactEmail}`"
+          :link-text="contactEmail"
+          :label="$t('Email')"
+          icon="mdi-email"
+        />
+        <icon-link
+          href="https://organise.earth/xr-netherlands/messages/@vacancies_support_xrnl"
+          link-text="@vacancies_support_xrnl"
+          :label="$t('Mattermost')"
+          icon="mdi-message"
```
SamLubbers, line 19: the Mattermost id should also come from `@/constants/contacts` just like email. So that if we change the contact in one place it changes everwhere

src/views/SupportPage.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+  <div class="container container--md container--margin-top">
+    <div class="row">
+      <div class="row__image">
+        <img src="@/assets/images/tree.svg" alt="Bee" />
```
SamLubbers, line 5: wrong alt text ;)

src/assets/images/close.svg
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SamLubbers, line 1: I see you've added a couple of standard standard `svg` files (close and hamburger). I think you can achieve the same by using the [Vuetify Icon component](https://vuetifyjs.com/en/components/icons/). Is this correct? Or is there and advantage to using our own files?

SamLubbers (changes requested):
Hey Tom, thanks again for creating this amazing new look for the navbar. I'm
very impressed how you've crated this completely from scratch with scss :astonished: :clap: Here I've shared some feedback about the design, and in the code I've left some comments and questions about the code.

### Design feedback

#### Issues

- On desktop the navbar should stick to the top. As you can see in the image,
  when you scroll down the list of roles the navbar also goes up. This is an
  issue because the navigation buttons are hidden and the filters section is
  left floating.
  
Screenshot from 2021-04-27 11-54-02

  
- On mobile the navbar does stay fixed, but when you click on filter, the filter
  drawer is hidden behind the navbar. The way I think it should work, and how it
  worked before, is that the filter view would appear on top of the navbar (and
  everything else visible in that page)

image


#### Improvements

- I think we should remove XR NL from the navbar and instead link to it from the
  About page. I think we should do that for 2 reasons:
  - Predictability: currently when you click on the navigation buttons you
  smoothly go from one page to the other, and when you click on XR NL you jump
  to a completely different website
  - Context: If we link to the xr nl website from the about page we can give the
  user some more information about the relation of both websites. For example,
  we could have something like: "this is a project started by [Extinction
  Rebellion NL](https://extinctionrebellion.nl/en). In the website you can learn
  XYZ. This apps tries to do XYZ"
- I think we can integrate the language toggle and menu hamburger better into
  the new navbar design by having the same background colour and font as the
  other buttons. Just like the [global XR website](https://rebellion.global/)
- In the last meeting we agreed that the default XR green is too strong on the
  eyes, and that instead we would use the lighter variation used by the global
  website: So `#4ea648` instead of `#14aa37`

src/styles/global.scss
SamLubbers: Before we achieve this by having the navbar height defined in the `Vuex.store`, so that other components could also make use of this variable to set their height (for example, the filter drawer makes use of this). It might make more sense to just define a `scss` variable, which I think can also be used from any component.

When (or if) you make this change, you'll have to adapt one unit test for the navbar (which is currently failing). The only test for the navbar is: `has height defined in store` (in `tests/unit/components/components.spec.js`).

In the readme you can see how to run the tests :v:

SamLubbers: And... one last thing. Since you're already fixing the tests, you can also fix this simple issue raised by Prettier:

```
aginun/src/components/TheAppBar.vue
  77:17  error  Delete `,`  prettier/prettier

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

I could have done this myself but it would have re-triggered the entire CI pipeline :sweat_smile:

package.json
tomootes: Thanks for the advice! I added it to my VSCode. Thats a lot easier

tomootes:
@SamLubbers this is my reply on the design part: 
- On desktop the navbar should stick to the top.
> I changed this to fixed. Now it will stay on top.

- The filter drawer falls behind the navbar.
> I suggest that we create a total new styling for the filter drawer according to the designs. I made a new issue for this: #156. When i fixed this, this issue is nonexistent anymore. I won't fix it in this PR then.

- I think we should remove XR NL from the navbar and instead link to it from the
About page.
> Sure!  I fixed it to some extent.

- I think we can integrate the language toggle and menu hamburger better into
the new navbar design by having the same background colour and font as the
other buttons. Just like the global XR website.
> Uhuh, i agree. But maybe make a new issue for this?

- In the last meeting we agreed that the default XR green is too strong on the
eyes. 
> Yes changed it.

- We could give the page title (on the left of the navbar) the same font style as the other buttons.
> Since its the title i would prefer to give it just a different look? I think the current look is suitable for that :)

tomootes:
@SamLubbers lets talk about the rest tomorrow in person 🦖

src/i18n/messages/nl.json
tomootes: Yes i added them manually! I assumed that this was the way as soon as i found the messages file. I think documentation would be very useful here.

src/router/index.ts
tomootes: I think its quite descriptive this way! Were also importing them from files with these filenames. And i think its logical to name the files after what they are :)

src/styles/base/_typography.scss
tomootes: I wanted to create a mixin that can be used application-wide. So i think font-heading is better suitable for that!

src/styles/global.scss
tomootes: As we discussed a little while ago, i assumed that on the long run maybe we could do without Vuetify and only use a couple of components that are useful to us? In this case its useful to have our own classes for this. Otherwise we would have to agree upon using Vuetify classes everywhere.

src/styles/global.scss
tomootes: The navbar now has a dynamic height which consists of `padding` and the height of the logo image. This way the padding will always be the same and the component as a whole is dynamic. 

I added a dynamic height in `variables.scss` for the navbar: `$nav-bar-height: calc(#{$logo-height} + #{$navbar-padding-vertical});` Is this a suitable solution for you?

tomootes: I added new unit tests for the navbar. First time in my life hahaaaaaaaaa

src/styles/variables.scss
tomootes: See previous comment about outphasing Vuetify :) This is just a decision.

src/views/AboutPage.vue
tomootes: Yup. Removed it!

src/views/SupportPage.vue
tomootes: Check!

src/views/SupportPage.vue
tomootes: Check, i added it to the constants

src/views/SupportPage.vue
tomootes: Check!

src/assets/images/close.svg
tomootes: I can't distinguish which way should be favoured. I used icons from thenounproject.com because in my opinion they looked a bit better than the standard icons from Vuetify :) The advantage that we could see is that we could do without Vuetify in the end

tomootes:
@SamLubbers i pushed my new changes :)

SamLubbers:
@tomootes thanks for all your changes and comments :raised_hands: :raised_hands: :raised_hands: I had time to review your other pull requests but not yet this one. I'll get back to it asap.

In the meantime you can have a look at my review of #153. Have a good week!!

src/styles/global.scss
aleesaan: I agree with Sam here, I think that **for now** we should use Vuetify classes as much as we can, rather than start adding custom classes/styles on top of the Vuetify ones. This is for a few reasons:

1. Creating and using custom styles will increment complexity, potential of conflicts with other (e.g. vuetify) styles, and increase the amount of code to maintain;
2. (related to the last point above) Once we will start moving away from Vuetify, we would have to also remember to check and remove any other (unused) custom class / styles we added;
3. Using existing classes creates less random issues with different browser/devices (as they're usually already handled properly by Vuetify).

tomootes: @aleesaan i removed the vanilla .container and .row classes here.

tomootes:
@aleesaan / @SamLubbers i addressed al the questions here. Maybe we can merge this PR?

src/styles/global.scss
SamLubbers: @tomootes yes this solution works! So now if we ever change the logo height or navbar padding, then the navbar height will also be adjusted accordingly.

And congrats with your first unit test :partying_face: pretty cool right? :smirk:

src/assets/images/close.svg
SamLubbers: thenounproject.com looks great :smile: 

And yes, we can keep these icons as we've decided to move away from Vuetify :ok_hand:

SamLubbers (dismissed):
@tomootes thank you for all fixing the last remaining issues :fist_raised: This PR can now be merged :arrow_up: 

I really appreciate that you replied to each of my individual comments on the code. I've replied in turn to each of them and marked all conversations as resolved :heavy_check_mark: 

It was also really helpful that you commented on each of my points by quoting my text. This really made it much easier for me to do the review and see what still needs to be done :handshake: I'll do the same in turn:

> Sure! I fixed it to some extent.

Amazing! Having these default about and support page already make the application so much more complete. Soon we'll fill those pages with the information specified in the new designs.

> Uhuh, i agree. But maybe make a new issue for this? [for the language selector]

Will do :ok_person:

SamLubbers:
I'll rebase from the master branch to resolve the merge conflicts

SamLubbers (approved):

PR #153, opened 2021-04-26
Feature/authorization hasura 130

### LINKED ISSUE #130: Authorization in Hasura
At the moment, there's no protection on the resources (roles/tasks) in the Hasura backend. 

## Feature description
Logged-in users should be able to add, edit and remove roles in Hasura. This should not be possible for logged-out users.

Describe a step-by-step use case of your feature. 

1. User can't add/edit/remove a role by sending a request to the backend
2. User logs in and receives an access token
3. User can now add/edit/remove roles by sending a request to the backend, including the access token. Hasura gets the token, and if it's valid the user can make changes on the resources

## Additional comments
First we keep it simple: any logged-in user can edit any resource. The next step will be to make sure users can only edit the resources that they themselves created.

tomootes:

- Apollo now longer uses the admin secret
- Set right permissions in Hasura for editing / inserting new roles 
- Added bearer token to Apollo when it is set
- Hasura now lets unauthorized users request `config`,`language`,`localGroup`,`role` and `workingCircle`
- Hasura now lets unauthorized users use the rolesSearch function

Storing the admin secret is a bad idea. Anyone can get the secret from the source and access our Hasura back-end. Therefore we now use JSON web tokens provided by Keycloak.

Resolves #130 

## Additional comments

More information on the authentication mechanism will need to be added to the [documentation](https://wiki.extinctionrebellion.nl/).


tomootes:
In the Hasura docker-compose.yml a new param was added: `HASURA_GRAPHQL_UNAUTHORIZED_ROLE: public`. This enables us to assign permissions this role. When a table was made available to this role, the data becomes publicly available. 

The following tables were made available publicly: `config`,`language`,`localGroup`,`role`,`workingCircle`. This was done by assigning the select permission to the `public` role.

The users who are logged in automatically get the `user` role in Hasura. I granted the `user` role the following rights: `insert`, `select` and `update` on the `role` table. This enables them to change and create records in the table.

SamLubbers:
@tomootes I just merged the branch `login-modal`, so I've updated this branch to be merged into `master` instead.

SamLubbers:
@tomootes because you've branched off Ale's login-modal branch, and because he made some changes to that branch before merging to master, I've rebased from master so that your new commits are applied on top of the latest commits from master (which include Ale's new code)

src/plugins/vue-apollo.ts
```diff
@@ -13,23 +13,27 @@ Vue.use(VueApollo);
...
   headers: {
-    "x-hasura-admin-secret": process.env.VUE_APP_API_KEY
+    // "x-hasura-admin-secret": process.env.VUE_APP_API_KEY
   }
 });
 
-const languageLink = setContext((_, { headers }) => {
+// set request headers based on current application state
+const dyanmicLink = setContext((_, { headers }) => {
+  const loginToken = Vue.$cookies.get('loginToken');
```
SamLubbers, line 22: Would it be possible to get the token from the store? (`token` state variable in `src/store/modules/user.ts`). That way you get the token from memory instead of from the browser. I think the role of the cookie is to keep users logged in after they log in a first time, even if they leave the page.

src/plugins/vue-apollo.ts
```diff
@@ -13,23 +13,27 @@ Vue.use(VueApollo);
...
   return {
     headers: {
       ...headers,
-      "x-hasura-lang": i18n.locale
+      "x-hasura-lang": i18n.locale,
+      // If the loginToken exists, this property will bed added
+      ...(loginToken && {authorization: `Bearer ${loginToken}`})
```
SamLubbers, line 28: How does Hasura know that the token is a valid from our authentication server? Is there any link between Hasura and Keycloack? In other words, could someone use a fake token to have full authorization access to the sever?

SamLubbers (changes requested):
@tomootes amazing that you got this to work so elegantly and so quickly :star2: 

I just reviewed the functionality and it works almost perfect. Here, and in the code, I share some issues, improvements, and questions:

### issues

- [ ] when I delete a role through the UI it isn't deleted. I get the console error: `Error: GraphQL error: missing session variables: "x-hasura-user-id"`. Creating and editing roles works well. 
- [ ] when I try to publish, edit or delete a role without the login token, I don't get any feedback about the operation having failed, and instead I get the following error in the console `Uncaught (in promise) Error: GraphQL error: no mutations exist`. At first I thought this wasn't an issue, because people shouldn't be able to do these operations if they're not logged in, but it could still be useful to add some kind of excepction handling & feedback to the user in case we ever run into issues because we haven't configured Hasura or Keycloack right. 
  - If you want to test making requests without authenticating, simply go to `src/store/modules/user.ts` and set `token` to any value. You'll be able to access all the functionality for logged in users without being actually logged in.

### improvements

- [ ] now that the Hasura admin secret isn't needed anymore, there are a couple of documents to update:
  - README.md (find places where we mention the api key and remove it).
  - SUPPORT.md (remove mentions of the api key)
  - also I think you can remove the commented line from `Auth`

SamLubbers:
@aleesaan Prettier raises no issues in my local setup but it does raise issues in the lint step of the CI pipeline. How does this happen? Do we need to upgrade the version of prettier in the pipeline or locally? :thinking:

aleesaan:
@SamLubbers strange, cause it does show them for me:

image

If I save the file, they're automatically fixed:

image

Maybe double check your setup, and @tomootes can you do the same? So that we won't get these lint error in the pipelines, as they'll be solved automatically when you save files. You can ask me in private if you need help troubleshooting!

SamLubbers:
@aleesaan thanks for checking this out :pray: I did `npm install` and the Prettier issues showed up when running `npm start` :smiley: 

The reason they didn't show up when editing the file is because I recently changed my setup and forgot to enable `ESLint`. I'll configure this asap so I don't push any more code with linting errors :innocent:

tomootes:
@SamLubbers thanks for thoroughly checking 👍  I'll go through the issues here:

> - Error when deleting a role

I fixed this by changing a permission in Hasura. I also added a catch for when an error occures, this is now visible in the interface.

> - When I try to publish, edit or delete a role without the login token, I don't get any feedback about the operation having failed.

I dove into this. First i changed the code to actually showing an error in the interface. This works now. The second thing i did was doing some Googling on the error. I concluded that the error that you are talking about is right for this situation: `GraphQL error: no mutations exist`. I must say that it is quite bad at describing what the matter is. What it is trying to say is that for this user rule, there are no mutations and thus the user can not edit the table. I hope this issue is ok for now. Could you check the checkbox if you agree?

About the package-lock.json. As soon as you install a new dependency the package.json changes and should be committed. It stores the exact versions of the libs. If we keep doing this it should be up-to-date?

src/plugins/vue-apollo.ts
tomootes: What is your consideration to get it from the store instead of the cookie? Im not sure if memory is faster than a cookie? I guess when we use the cookies the code would be more descriptive in itself?

src/components/roles/RoleEditDialog.vue
```diff
@@ -339,19 +338,33 @@ export default {
...
+        this.createRole(role)
+          .then((succes) => {
+            this.displaySuccess(this.$t("Role created"));
+          })
+          .catch((error) => {
+            this.displayError(this.$t("An error occured creating this role"));
+          });
       }
+    },
+    onSucces() {
```
aleesaan, line 362: When I create/edit a role, the `RoleEditDialog` doesn't close automatically, and I see that it's probably because this isn't being used. Can you have a look?

aleesaan (changes requested):
I also finally had some time to look at your code, I left some of comments in the code but here's a few more things:

1. I just ran `npm i` on master and I don't get any changes in my package-lock, whereas when I run `npm i` on your branch I do get some changes (I guess the differences with the master one). Can you try running `npm i` again on your branch, and then push the package-lock changes? I think it will revert the current changes and bring back the 'master version'.  
In general, we should only update the package-lock when we update the dependencies, to make sure that everyone who `npm install`s always gets the same versions.
2. Can you remove the `VUE_APP_API_KEY` from the .env.example file, now that it's not used anymore?
3. If I login, then refresh the page, I won't get any roles. Can you have a look?

src/plugins/vue-apollo.ts
tomootes: Alright. I agree that in terms of consistency this is a good idea. I added it to the PR :)

README.md
```diff
@@ -24,8 +24,7 @@ You can install the application in your own computer by following these simple s
 
 1. Clone the repository by running `git clone https://github.com/xrnl/aginun.git` in your terminal.
 2. Navigate into the project directory (`cd aginun`) and install packages with the command `npm install`
-3. [Request the development API key](https://github.com/xrnl/aginun/blob/master/SUPPORT.md).
-4. Copy the `.env.example` file to `.env.local`, as follows: `cp .env.example .env.local` Once you have the API key, replace `<API_KEY>` in the `.env.local` file with your API key.
```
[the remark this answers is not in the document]
tomootes: I do agree! I added it again

src/components/roles/DeleteRoleConfirmation.vue
```diff
@@ -40,12 +40,19 @@ export default {
...
     ...mapActions("roles", ["deleteRole"]),
-    ...mapActions("alerts", ["displaySuccess"]),
+    ...mapActions("alerts", ["displaySuccess", "displayError"]),
     onDeleteRole() {
-      this.deleteRole(this.roleId);
-      this.$emit("input", false);
-      this.displaySuccess("Role deleted");
-      this.$router.push("/roles");
+      this.deleteRole(this.roleId)
+        .then((succes) => {
```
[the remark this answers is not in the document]
aleesaan: If you have autofix enabled it will pick pick up warnings as well, but not this kind lint errors - we need to resolve these ourselves. Easiest way is to keep an eye on the terminal and make sure that there are no warnings there when compiling the code!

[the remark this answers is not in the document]
tomootes: Sure. I removed it 👍

src/components/roles/DeleteRoleConfirmation.vue
```diff
@@ -40,12 +40,19 @@ export default {
...
+          this.displaySuccess("Role deleted");
```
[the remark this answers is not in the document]
tomootes: > I also finally had some time to look at your code, I left some of comments in the code but here's a few more things:
> 
> 1. I just ran `npm i` on master and I don't get any changes in my package-lock, whereas when I run `npm i` on your branch I do get some changes (I guess the differences with the master one). Can you try running `npm i` again on your branch, and then push the package-lock changes? I think it will revert the current changes and bring back the 'master version'.
>    In general, we should only update the package-lock when we update the dependencies, to make sure that everyone who `npm install`s always gets the same versions.

When i run `npm install` on my branch it doesn't create any changes in de package-lock.json. A while ago the whole project lacked any package-lock.json so thats why i added mine on a certain point in time. Maybe we can go with this one?

> 2. Can you remove the `VUE_APP_API_KEY` from the .env.example file, now that it's not used anymore?

Yesss

> 3. If I login, then refresh the page, I won't get any roles. Can you have a look?

This is a bug that initiated when i introduced the communicating with Hasura using the bearer token. It's quite a hard one so maybe we can turn it into an issue. And close this PR :)

tomootes:
> I also finally had some time to look at your code, I left some of comments in the code but here's a few more things:
> 
> 1. I just ran `npm i` on master and I don't get any changes in my package-lock, whereas when I run `npm i` on your branch I do get some changes (I guess the differences with the master one). Can you try running `npm i` again on your branch, and then push the package-lock changes? I think it will revert the current changes and bring back the 'master version'.
>    In general, we should only update the package-lock when we update the dependencies, to make sure that everyone who `npm install`s always gets the same versions.

When i run `npm install` on my branch it doesn't create any changes in de package-lock.json. A while ago the whole project lacked any package-lock.json so thats why i added mine on a certain point in time. Maybe we can go with this one?

> 2. Can you remove the `VUE_APP_API_KEY` from the .env.example file, now that it's not used anymore?

Yesss

> 3. If I login, then refresh the page, I won't get any roles. Can you have a look?

This is a bug that initiated when i introduced the communicating with Hasura using the bearer token. It's quite a hard one so maybe we can turn it into an issue. And close this PR :)

src/components/roles/RoleEditDialog.vue
```diff
@@ -339,19 +338,33 @@ export default {
       };
 
       if (this.editRole) {
-        this.updateRole({ id: this.editRole.id, ...role });
+        console.log("heja");
```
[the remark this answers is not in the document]
tomootes: Haha, of course 🙈

src/components/roles/RoleEditDialog.vue
```diff
@@ -339,19 +338,33 @@ export default {
...
+        this.updateRole({
+          id: this.editRole.id,
+          ...role
+        })
+          .then((success) => {
```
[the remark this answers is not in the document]
tomootes: I checked out some other code and it looks a lot cleaner and clearer. Implemented it 🦅

src/components/roles/RoleEditDialog.vue
tomootes: I added it to the onSubmit function again. Forget to put it back there.

tomootes:
@aleesaan i added new commits according to your suggestions. Maybe you could check it one more time before we merge it?

README.md
```diff
@@ -23,9 +23,10 @@ If you want to get in contact or learn more about the project, please read the [
 You can install the application in your own computer by following these simple steps:
 
 1. Clone the repository by running `git clone https://github.com/xrnl/aginun.git` in your terminal.
-2. Navigate into the project directory (`cd aginun`) and install packages with the command `npm install`
-3. [Request the development API key](https://github.com/xrnl/aginun/blob/master/SUPPORT.md).
-4. Copy the `.env.example` file to `.env.local`, as follows: `cp .env.example .env.local` Once you have the API key, replace `<API_KEY>` in the `.env.local` file with your API key.
+2. Navigate into the project directory (`cd aginun`) and install packages with the command `npm install`.
+3. Copy the `.env.example` file to `.env.local`, as follows: `cp .env.example .env.local`.
+4. [Request the necessary credentials/ keys.](https://github.com/xrnl/aginun/blob/master/SUPPORT.md).
+
```
aleesaan, line 29: Can you remove this extra line?

src/components/roles/DeleteRoleConfirmation.vue
```diff
@@ -40,12 +40,19 @@ export default {
...
     ...mapActions("roles", ["deleteRole"]),
-    ...mapActions("alerts", ["displaySuccess"]),
+    ...mapActions("alerts", ["displaySuccess", "displayError"]),
     onDeleteRole() {
-      this.deleteRole(this.roleId);
-      this.$emit("input", false);
-      this.displaySuccess("Role deleted");
-      this.$router.push("/roles");
+      this.deleteRole(this.roleId)
+        .then(() => {
```
aleesaan, line 46: Can you use async/await here too, and also move the alert logic in the store module?

src/components/roles/RoleEditDialog.vue
```diff
@@ -318,7 +317,7 @@ export default {
   },
   methods: {
     ...mapActions("roles", ["updateRole", "createRole"]),
-    ...mapActions("alerts", ["displaySuccess"]),
+    ...mapActions("alerts", ["displaySuccess", "displayError"]),
```
aleesaan, line 320: We can delete this now.

src/components/roles/RoleEditDialog.vue
```diff
@@ -339,19 +338,30 @@ export default {
       };
 
       if (this.editRole) {
-        this.updateRole({ id: this.editRole.id, ...role });
+        try {
+          await this.updateRole({
+            id: this.editRole.id,
+            ...role
+          });
+          this.onSucces();
```
aleesaan, line 346: If something goes wrong during the role creation / update, we shouldn't call this method. Also, since we try/catch inside the store methods (and more specifically since they will never throw errors), we don't need to do it again here. We can do something like this here:

```js
const serverError = this.editRole
  ? await this.updateRole({
      id: this.editRole.id,
      ...role
    })
  : await this.createRole(role);

if (!serverError) {
  this.$emit("input", false);
  this.resetState();
  this.$nextTick(() => {
    this.$refs.form.reset();
  });
}
```

We can then drop the `onSucces` method here, and add a `return error` at the end of the `catch` in the update / create role (store) methods, so we can return the error to this component and assign it to the `serverError` variable.

src/store/modules/roles.ts
```diff
@@ -85,38 +86,64 @@ export default {
...
+          variables: { input: [newRole] },
+          update: (
+            store,
+            {
+              data: {
+                insert_role: { returning }
+              }
             }
+          ) => {
+            commit("addRole", returning[0]);
```
aleesaan, line 102: We can move the `dispatch("alerts/displaySuccess", ...)` here as well.

src/store/modules/roles.ts
```diff
@@ -85,38 +86,64 @@ export default {
...
+          update: (
+            store,
+            {
+              data: {
+                update_role: { returning }
+              }
             }
+          ) => {
+            commit("editRole", returning[0]);
+            dispatch("loadRoles");
```
aleesaan, line 132: Same here, let's move the alert after this.

aleesaan (commented):
Super nice to see you quickly implemented the feedback, and that you also fixed the tests! I have few more comments 😇   
Also, about this:

> When i run npm install on my branch it doesn't create any changes in de package-lock.json. A while ago the whole project lacked any package-lock.json so thats why i added mine on a certain point in time. Maybe we can go with this one?

I tried removing the node_modules folder completely, then do an `npm i`, and I still get changes in the package-lock file. Can you try doing the same? Then you can push the changes you get.

tomootes:
> Super nice to see you quickly implemented the feedback, and that you also fixed the tests! I have few more comments 😇

Testing is the future right? 🤖 

> Also, about this:
> 
> > When i run npm install on my branch it doesn't create any changes in de package-lock.json. A while ago the whole project lacked any package-lock.json so thats why i added mine on a certain point in time. Maybe we can go with this one?
> 
> I tried removing the node_modules folder completely, then do an `npm i`, and I still get changes in the package-lock file. Can you try doing the same? Then you can push the changes you get.

With removing node_modules this works for me as well! I committed the new package-lock.json.

README.md
tomootes: Alrighty then

src/plugins/vue-apollo.ts
SamLubbers: great! 

These past days I've indeed been learning how Hasura works with keycloack and how it exchanges information through encrypted JWT tokens :ok_hand:

SamLubbers (approved):
@tomootes  I finally got around to making the final review of this pull request :muscle:

- The error handling now works well! If a user doesn't have permissions to a query/mutation or if they don't have permissions for a certain operation, they get some feedback indicating that something went wrong :ok_hand:
- Now when reviewing this PR for the second time I realised that there was still some small changes to make in the documentation. I've made these changes myself and merged them into master.
- @aleesaan I've also incorporated all your final suggestions to call the alerts from the roles Vuex store, and to replace all then/catch with async/await + try/catch.

I will approve & merge this change into the master branch because to fix #159 I had to make a change in the backend that breaks the current code in the master branch. By merging now we will have a working master branch. If there is anything I messed up in these latest changes let me know and I'll fix it :pray:

PR #157, opened 2021-05-03
Feature/new layout filter 156

### LINKED ISSUE #156: New layout filter on mobile
Right now the filter bar isn't that usable on mobile. This should be changed to work with the navbar.

tomootes:

**Before**: the button hangs still at the top:
Schermafbeelding 2021-05-03 om 12 02 59

**After**: the button hangs at the bottom:
Schermafbeelding 2021-05-03 om 11 50 29

**Before**: here the back button in the filter is difficult to find:
Schermafbeelding 2021-05-03 om 12 03 14

** After **: here the back button is located at the bottom and easier to find:
Schermafbeelding 2021-05-03 om 11 50 38

- It was very unclear for users how to filter the search results.
- Both buttons were located at the top, which is harder to tap for mobile users.

Resolves #156 

## Additional comments

Quite straightforward!

### DIFF
--- a/src/components/AutocompleteCustom.vue
+++ b/src/components/AutocompleteCustom.vue
@@ -1,13 +1,12 @@
 <template>
-  <v-autocomplete
+  <v-select
     :items="items"
     :value="selectedItemsIds"
     item-value="id"
     :item-text="itemText"
-    chips
-    small-chips
-    deletable-chips
     multiple
+    outlined
+    :hide-details="hideDetails"
     class="mt-3"
     :label="label"
     @change="$emit('change', $event)"
@@ -20,6 +19,10 @@ import { isArrayValid } from "@/utils/validators";
 export default {
   name: "AutocompleteCustom",
   props: {
+    hideDetails: {
+      type: Boolean,
+      required: false
+    },
     items: {
       type: Array,
       required: true
--- a/src/components/AutocompleteCustom.vue
+++ b/src/components/AutocompleteCustom.vue
@@ -1,13 +1,15 @@
 <template>
-  <v-select
+  <v-autocomplete
     :items="items"
     :value="selectedItemsIds"
     item-value="id"
     :item-text="itemText"
-    multiple
     outlined
+    chips
+    small-chips
+    deletable-chips
+    multiple
     :hide-details="hideDetails"
-    class="mt-3"
     :label="label"
     @change="$emit('change', $event)"
   />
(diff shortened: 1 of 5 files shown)


tomootes:
Im sorry but some changes in this feature branch were meant for resolving #130. Maybe we can resolve this one first after reviewing #130 again.

SamLubbers (commented):
@tomootes these are great changes :raised_hands: See the [comments on Mattermost](https://organise.earth/xr-netherlands/pl/trf9f5qysfrb9cqdbgxf9ymf9o) for feedback from me and Ale. Next time you want some feedback on a pull request you can just send a link to that pull request on Mattermost. That way we don't duplicate conversations across platforms.

And which changes were fixes intended for #153? If you want we can already merge this branch into that one so that it has all of your latest fixes.

SamLubbers (changes requested):
@tomootes thank you so much for implementing the feedback so quickly :zap: The new design is great!! :fire: 

There is only a couple of small things that I would change and that can be done quite quickly:

1. Allow users to select multiple local groups or working groups. Before we used `v-autocomplete` so that users could find roles within multiple local groups or working groups, which is very useful in case someone is interested in roles in more than one group. I think that with these changes (using `v-select`), users can only select one group at a time.
2. Disable (but do not hide) the "remove filters" button: this is more a suggestion than a requirement. I think that if a certain button cannot be used, we should still show it and disable it, instead of hiding it completely. I think this makes the interface more consistent and provides less surprises.

tomootes:
> @tomootes thank you so much for implementing the feedback so quickly ⚡ The new design is great!! 🔥
> 
> There is only a couple of small things that I would change and that can be done quite quickly:
> 
> 1. Allow users to select multiple local groups or working groups. Before we used `v-autocomplete` so that users could find roles within multiple local groups or working groups, which is very useful in case someone is interested in roles in more than one group. I think that with these changes (using `v-select`), users can only select one group at a time.
> 2. Disable (but do not hide) the "remove filters" button: this is more a suggestion than a requirement. I think that if a certain button cannot be used, we should still show it and disable it, instead of hiding it completely. I think this makes the interface more consistent and provides less surprises.

hi @SamLubbers its been a pleasure man 🌱

SamLubbers (approved):
@tomootes WOHOOO this branch is now ready to be merged :tada: I'll rebase from master and skip the commits included in #152 in order to resolve the merge conflicts :sunglasses:

Issue #155, opened 2021-05-03
New layout role cards


tomootes:
Because the figma design differs from our current role card designs, we want to change the styling/ layout a bit.

The layout should look according the new designs:
![Schermafbeelding 2021-05-05 om 11 41 52](https://user-images.githubusercontent.com/4222167/117123129-f01c0580-ad96-11eb-9245-bdb84586977d.png)

PR #158, opened 2021-05-05
New styling for role cards

tomootes:

Cards desktop before:
Schermafbeelding 2021-05-05 om 16 47 51

Cards desktop after:
Schermafbeelding 2021-05-05 om 16 46 00

Cards mobile before:
Schermafbeelding 2021-05-05 om 17 22 10

Cards mobile after:
Schermafbeelding 2021-05-05 om 17 21 27

Clearer, lighter overview of all roles.

#155 

## Additional comments

Pretty straightforward. I hope its ok to use props instead of slots. I think this is ok for now since we're not implementing very different types of cards anytime soon.

I removed the inline styling of the gridList as well. In my opinion it was unnecessary complicated.


SamLubbers:
Oh and one last thing, does the new design support very long titles (I think it's max 20/30 characters on the form to create a new role)

tomootes:
@SamLubbers thanks! :) 

These are my two cents when it comes to the questions:
- The name of the working group. I'm not sure if we need these on the cards. Also, i would like to hear this from the people who need to work with the platform. Maybe we can test the platform as well on to-be rebels?
- Im not sure about the second point.
- When it comes to the read-more button, i think the cards are really obvious to click on. Especially with the shadow.
- About the long titles, i added a fix for this to stay inline. This now looks like this:

Schermafbeelding 2021-05-10 om 09 27 51

Also! If you consider this version ok, maybe we can we can merge this PR?

SamLubbers (approved):
@tomootes @kajdehoop thank you so much for your input!

- keeping it like this and testing it with users sounds good to me
- I'm also happy without read more
- Tom thanks for adding some code to deal with long titles, and Kaj we do have a character limit so that people don't go crazy on the role titles ;)

this PR is now... OFFICIALLY READY TO GO :rocket: :rocket: :rocket:

PR #163, opened 2021-05-27
Filters in URL

aleesaan:
#26

### DIFF
--- a/src/components/roles/RoleViewDialog.vue
+++ b/src/components/roles/RoleViewDialog.vue
@@ -12,8 +12,8 @@
       persistent
       max-width="750"
       value="true"
-      @click:outside="$router.push('/roles')"
-      @keydown.escape="$router.push('/roles')"
+      @click:outside="$router.back()"
+      @keydown.escape="$router.back()"
     >
       <v-card>
         <transition name="fade" mode="out-in">
@@ -290,7 +290,7 @@ export default {
       const error = await this.fillRole(this.role.id);
       if (!error) {
         this.$emit("input", false);
-        this.$router.push("/roles");
+        this.$router.back();
       }
     },
     formatDate(date) {
--- a/src/router/index.ts
+++ b/src/router/index.ts
@@ -1,7 +1,7 @@
 import Vue from "vue";
 import VueRouter from "vue-router";
 
-import { rolesErrorGuard, hasErrorsGuard } from "./guards";
+import { rolesGuard, hasErrorsGuard } from "./guards";
 import ErrorPage from "../views/ErrorPage.vue";
 import RolesOverview from "../views/RolesOverview.vue";
 import AboutPage from "../views/AboutPage.vue";
@@ -14,9 +14,8 @@ const routes = [
   {
     path: "/roles",
     name: "roles",
-    beforeEnter: rolesErrorGuard,
+    beforeEnter: rolesGuard,
     component: RolesOverview,
-    alias: "/",
     children: [{ path: "view/:id", component: RoleViewDialog }]
   },
   {
@@ -36,9 +35,13 @@ const routes = [
     component: SupportPage
   },
   {
+    path: "/",
+    redirect: "/roles"
+  },
+  {
     // non-existent pages redirect to the home page
     path: "*",
-    redirect: "/"
+    redirect: "/roles"
   }
 ];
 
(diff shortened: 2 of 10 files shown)


src/router/index.ts
```diff
@@ -35,10 +34,14 @@ const routes = [
     name: "support",
     component: SupportPage
   },
+  {
+    path: "/",
+    redirect: "/roles"
+  },
```
SamLubbers, line 40: Nice that you're redirecting to roles instead of using an alias. This makes the URL behaviour consistent and predictable

src/components/roles/RoleViewDialog.vue
```diff
@@ -12,8 +12,8 @@
       persistent
       max-width="750"
       value="true"
-      @click:outside="$router.push('/roles')"
-      @keydown.escape="$router.push('/roles')"
+      @click:outside="$router.back()"
+      @keydown.escape="$router.back()"
```
SamLubbers, line 16: Great that you've fixed this by using the correct Vue Router function :raised_hands: Using the Vuex URL stack properly will avoid some nasty errors in the future :100:

src/styles/components/_nav-bar.scss
```diff
@@ -70,7 +70,7 @@
         position: fixed;
         top: 0;
         left: 0;
-        background: $xr-green;
+        background: $xr-green-light;
```
SamLubbers, line 73: I also noticed this last time hehe. Good that you fixed it :sunglasses:

src/store/modules/roles.ts
```diff
@@ -266,18 +283,28 @@ export default {
...
     },
     setDefaultFilters({ commit, dispatch }) {
       commit("setLoadingState", true);
-      commit("setFilter", { filterType: "search", filterValue: "" });
-      commit("setFilter", { filterType: "localGroups", filterValue: [] });
-      commit("setFilter", { filterType: "workingCircles", filterValue: [] });
-      commit("setFilter", {
-        filterType: "timeCommitment",
-        filterValue: [timeCommitmentRange.min, timeCommitmentRange.max]
-      });
```
SamLubbers, line 279: What an amazing simplification of the code :sparkles: :100:

tests/unit/store/roles.spec.ts
```diff
@@ -411,22 +425,7 @@ describe("Roles Store", () => {
...
-        });
-        expect(commit).toBeCalledWith("setFilter", {
-          filterType: "workingCircles",
-          filterValue: []
-        });
-        expect(commit).toBeCalledWith("setFilter", {
-          filterType: "timeCommitment",
-          filterValue: [timeCommitmentRange.min, timeCommitmentRange.max]
-        });
+        expect(commit).toBeCalledWith("setDefaultFilters");
```
SamLubbers, line 428: Once again, amaaazing that you've also implemented tests for your changes (and simplified the tests so much)

SamLubbers (changes requested):
@aleesaan what a well-crafted clean change :star2: (as always :wink:) Throughout the code I've sprinkled some specific tokens of appreciation.

I've only identified one issue and one question

### Issues 

The working group and local group filters in the URL should use the group name instead of the internal ID in the database because this ID might change. If we use the id people might get different results if something in the database changes, but if we use the group name then they'll always get the results for the groups they want to filter by.

### Questions

Currently when I enter an invalid URL parameter it shows all roles (as it should), but it keeps the parameter in the URL. Is this what we want? Or do we want to remove invalid parameters from the URL?

aleesaan:
Awesome to hear 🚀  
To answer your points:

> ### Issues
> The working group and local group filters in the URL should use the group name instead of the internal ID in the database because this ID might change. If we use the id people might get different results if something in the database changes, but if we use the group name then they'll always get the results for the groups they want to filter by.

I disagree, we should always rely on IDs for this stuff, the name isn't reliable for two reasons: it can change in different languages (e.g. den Haag vs the Hague for a local group), and it could change over time (e.g. a working group changing name). The database's IDs should never change, if anything we'd add more groups into a new ID; but even if it did, the change would be reflected in the filters, so the correct group would be selected!

> ### Questions
> Currently when I enter an invalid URL parameter it shows all roles (as it should), but it keeps the parameter in the URL. Is this what we want? Or do we want to remove invalid parameters from the URL?

Good idea! I added the logic to clear the extra params in my last commit ✌️

SamLubbers (approved):
@aleesaan good point about the names of groups in different languages! I didn't consider that one. I'm sold :wink:

PR #166, opened 2021-05-31
Fix production credentials

### LINKED ISSUE #154: Do not load user login credentials in production mode

1. Execute the command `npm run build` to build the production version of the app.
2. start app by visiting the generated index.html file
3. click on the log in button 

There is no data in the usernmae or password fields. The user must enter this information.

## Actual result

The username and password of the test user is entered.

## Additional comments

In development mode (`process.env.NODE_ENV !== "production"`) the default user data should still be loaded so that developers can easily log in while developing the application. This data should just not be loaded in production. 

The relevant file for this change is `src/components/AuthState.vue` Note that these credentials shouldn't only not be loaded in `data` (which determines the data on first load) but also in `resetForm()` (which determines the data when the user has logged out.

aleesaan:

Resolves #154

### DIFF
--- a/src/components/AuthState.vue
+++ b/src/components/AuthState.vue
@@ -87,6 +87,7 @@ import { mapGetters, mapActions } from "vuex";
 import { ValidationProvider, ValidationObserver } from "vee-validate";
 import { PulseLoader } from "@saeris/vue-spinners";
 import { contactEmail } from "@/constants/contacts";
+import { isProduction } from "@/utils/isProduction";
 
 export default Vue.extend({
   name: "AuthState",
@@ -94,8 +95,8 @@ export default Vue.extend({
     return {
       contactEmail,
       dialog: false,
-      username: "kaj-dev",
-      password: "test",
+      username: !isProduction() ? "kaj-dev" : "",
+      password: !isProduction() ? "test" : "",
       serverError: "",
       serverLoading: false
     };
@@ -114,8 +115,8 @@ export default Vue.extend({
     ...mapActions("user", ["login", "logout"]),
     ...mapActions("alerts", ["displayError"]),
     resetForm() {
-      this.username = "kaj-dev";
-      this.password = "test";
+      this.username = !isProduction() ? "kaj-dev" : "";
+      this.password = !isProduction() ? "test" : "";
       this.serverLoading = false;
       this.serverError = "";
       (this.$refs.login_form as InstanceType<
--- a/src/utils/isProduction.ts
+++ b/src/utils/isProduction.ts
@@ -1,0 +1,3 @@
+export function isProduction() {
+  return process.env.NODE_ENV === "production";
+}
--- a/src/views/ErrorPage.vue
+++ b/src/views/ErrorPage.vue
@@ -23,12 +23,13 @@
 
 <script>
 import { contactEmail } from "@/constants/contacts";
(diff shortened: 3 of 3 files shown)


src/components/AuthState.vue
```diff
@@ -87,15 +87,16 @@ import { mapGetters, mapActions } from "vuex";
...
 
 export default Vue.extend({
   name: "AuthState",
   data() {
     return {
       contactEmail,
       dialog: false,
-      username: "kaj-dev",
-      password: "test",
+      username: !isProduction() ? "kaj-dev" : "",
```
jurcello, line 98: Could you move these values to the env file? Might be better to leave credentials out of the code.

aleesaan: Yea I had thought about it, but I think this is better, since anyway we'd have to put them in the `.env.example` file - or people would always need to ask us the test credentials when they'd like to try running the project locally. These credentials won't work on production anyway, so it's fine to keep them here.

jurcello: I can agree with that :-)

src/components/AuthState.vue
```diff
@@ -87,15 +87,16 @@ import { mapGetters, mapActions } from "vuex";
 import { ValidationProvider, ValidationObserver } from "vee-validate";
 import { PulseLoader } from "@saeris/vue-spinners";
 import { contactEmail } from "@/constants/contacts";
+import { isProduction } from "@/utils/isProduction";
```
jurcello, line 90: Is it the intention to add more utility functions to this file? Otherwise a default export is also an option.

aleesaan: Not really, but I prefer named exports over default ones for a few reasons:

1. They're more flexible as they can be used to export multiple things from the same file without the need for mixed syntax like `import alertsStore, { AlertsState } from "@/store/modules/alerts"`;
2. They provide more structure as you (almost) always import using the defined variable name, so it's easier when you want to do a quick search through the code;
3. They make it easier for intellisense to import stuff, as you can start typing the variable name and it automatically completes and imports it for you.

Of course the above aren't absolute, both type of imports work fine, but yea I prefer named exports :)

SamLubbers:
@jurcello you were still going to review these changes, right?

src/components/AuthState.vue
jurcello: I understand your reasoning but I don't totally agree with all the reasons you give. I use webstorm and there the autocomplete also works with a default export. I would prefer a default export if there is only one function, so it's clear there are no other functions in that file (the store export is also a default export for instance).

On the other hand, I don't care too much either, so here it's completely fine with me. I would however suggest that we create a coding standard for exports, so at least in our own code we always use it consistently.

jurcello:
I reviewed the changes and I think this branch can be merged.

Based on this pull request I would suggest we agree on a coding convention how to do exports when there is only one function in a file, so we have consistency in the code.

SamLubbers:
@jurcello good idea to have these sort of things documented. Could you make a new page in the wiki for coding conventions? You could add the import convention as the first point :wink: 

And good to hear that the PR is approved :heavy_check_mark: :raised_hands: Next time it would be useful if you could approve it on Github so that Github knows it's approved and it allows us to merge this branch. You can do that by going to `files changed > review changes > approve > submit review`

SamLubbers (approved):

PR #213, opened 2021-06-26
Issue #178: Disabled apollo cache for all queries. Closes #178.

### LINKED ISSUE #178: Fix apollo caching

1. apply a filter that shows a subset of roles. For example, type something in the search field. 
2. delete one of the role
3. remove filter(s)

The role that we've deleted doesn't show anymore

## Actual result

The role that we've delete appears again after removing filters.

## Additional comments

This error occurs because Apollo has caching enabled. The first time you load all roles is stores the result for that query in a cache. After you delete the role, and then make the same query as the first time, it doesn't get all the roles from the backend, but instead gets the roles from the cached result of the query. To fix this we can either:

1. Remove apollo caching completely, and use the Vuex store as the single source of truth
2. Refresh the apollo cache every time a role is created/edited/deleted.

jurcello:

Disabled caching in apollo for all queries.

Resolves #178 

## Additional comments

Please explain difficult parts of your code to make it easier to review your pull request.

### DIFF
--- a/src/plugins/vue-apollo.ts
+++ b/src/plugins/vue-apollo.ts
@@ -33,7 +33,8 @@ const apolloClient = new ApolloClient({
   cache,
   defaultOptions: {
     query: {
-      errorPolicy: "all"
+      errorPolicy: "all",
+      fetchPolicy: "no-cache"
     }
   }
 });


SamLubbers (approved):
super slim & neat solution :clap: thanks for fixing this so quickly!

Issue #189, opened 2021-06-05
Set correct body font


kajdehoop:
Set body font to Lucida Sans 18px with color #605E5E ([Download font here](https://zeroheight.com/4eba4d3a1/p/175e89-downloads))

Issue #173, opened 2021-06-02
Implement design: About page


kajdehoop:
Main purpose of this page is explaining what XR is and how it is structured. The design of the interactive element explaining the structure is not final yet, might change in the future. The general layout of this page is finished however.

[Link to design](https://zeroheight.com/4eba4d3a1/p/063c65-about/b/407259)

SamLubbers:
I think part of this issue is specified in #138 You can decide which issue to close and which one to keep :ok_hand:

kajdehoop:
I'll close this one 👍

PR #214, opened 2021-06-26
Implement about page designs

aleesaan:
#173, #189 

## Additional comments
I haven't included the part about groups because we still need to finalize the discussion about how we want to show them.  
@SamLubbers and @kajdehoop could you have a look at the texts and let me know how to adjust them? I've taken them from the designs but they weren't complete there.


SamLubbers (changes requested):
@aleesaan awesomeeee changes!! :heart_eyes: you didn't only implement the about page, but also already sorted out some other important things like changing the font, adding illustrations & adding colours :sparkles: 

There are a few small things that I would change to make it consistent with the design:

- white border around each section. Note that the borders between sections should be just as thick as the border on the left and right.
- more padding on the x-axis (on desktop)
- add background icons

You can probably easy copy these styles from the [global page](https://rebellion.global/) hehe. Also, it would be amazing if you could make a reusable component for each section, so that if we want to add a new section we just need to pass the background-colour & background icon as props, and add the content (title, text & other html) through slots.

aleesaan:
Hey @SamLubbers sorry I had time to look into your feedback only now. The changes you mention would require quite a lot of extra work that would need to be redone after the bootstrap upgrade, so I think it's better not to spend too much time on these finishing touches just yet.

On top of that, I wanted to mention that I'm still not too convinced about this page. I think we have a bit too much info (that would need to be maintained+translated) that could be found in the XR global page, which in my opinion is a better place to read this kind of into.  
I suggest two options:

1. Make this about page more minimalistic and basically only link to the global site for more info;
2. Embed the https://rebellion.global/about-us/ page inside our about page via an iframe.

Anyway, if you agree for now I'd just update the texts to take away the lorem ipsum and make sure they make sense all together, and merge this branch. Then we can discuss my points above and if we decide to still go with the current designs, I will create a component and do the final adjustments after the bootstrap upgrade.

SamLubbers (approved):
@aleesaan good point! Let's keep it like this for now and make the last design improvements when we have Bootstrap. 

And yeah we want to keep this page as minimal as possible, and should definitely link to other sites for more information. Today I will:

* implement your first suggestion: remove some text and link to other XR websites
* add the missing text, which is currently lore ipsum

SamLubbers:
@aleesaan I've finally updated the about page :partying_face: Here is a summary of my changes, as well as a screenshot of the current version:

- I've updated the top section with a clear call to action: why they should help in XR and how they can best do that
- I've removed the structure section, and linked to the structure page of the XR NL website at the bottom
- I've added a section, explaining the platform, as well as a call to action to help with the platform
- I've deleted the final section `what if I can't find a good role`, and instead made this better suggestion #219 
- I've also not added instructions for how to publish a role, and instead made a new issue that incorporates this better into the user flow #218 

Screenshot 2021-07-20 at 13-02-10 Aginun

aleesaan:
Thanks for the changes @SamLubbers looks great 🙌

PR #247, opened 2021-11-18
configure backend permissions for editing and deleting roles

SamLubbers:
## What changes did you make? 

I've made the backend changes necessary so that users can only edit or delete their own roles:

1. I've changed the keycloack configuration so that it returns all the expected values by Hasura. In the content section of the jwt token you'll now find this for every user:

```json
{
  "hasura": {
    "x-hasura-default-role": "user",
    "x-hasura-user-id": "d295c4dd-f6c1-4d62-9213-27402823c150",
    "x-hasura-allowed-roles": [
      "user"
    ]
  },
}
```
To get the id of the user from the jwt token, you can get it from `jwt["hasura"]["x-hasura-user-id"]`, or from `jwt["sub"]` as keycloack automatically stores the id of the user in the `sub` field.

2. I've configured the backend so that whenever a new role is created, it stores the `x-hasura-user-id` (the id that the user has on keycloack` into the `authorId` column of the role
3. I've configured the backend so that users can only edit and delete their own roles.

@aleesaan I'm not sure if the permissions I configured in 2 & 3 are stored in the new init.sql file. You'll have to setup the db from scratch and see if it works as expected :crossed_fingers: 

If it works, the next step is to implement the frontend logic so that users don't see the edit and delete buttons for roles that aren't theirs. 

On top of that we'll have to fix some issues. I've already found these issues while testing:

- [x] on update hasura returns this error: `field "createdDate" not found in type: 'role_set_input'`. This is an issue that must have existed also before these changes
- [x] only delete role from local storage if the request is successful. Hasura always returns a 200 success code, even if no rows are deleted. We need to check the `affected_rows` property of the returned object. If `affected_rows = 1` then delete the role, otherwise show an error message.


aleesaan (commented):
I just tried deleting my local DB and resetting it up from scratch with your branch, but I'm getting this error:

image
  
  
> @aleesaan I'm not sure if the permissions I configured in 2 & 3 are stored in the new init.sql file. You'll have to setup the db from scratch and see if it works as expected 🤞
  
I think it would be best if you did this! Then you can see if it works as expected - if it doesn't, we'll need to figure out how to get it to work, as the initial docker setup should take care of the permissions as well.

SamLubbers:
@aleesaan thanks for the review :pray: 

Something clearly went wrong with storing and restoring the database. That function is missing even though it should be there. I'll look into it and fix it.

And you're right: it's better if i test the changes locally and make sure it works before sharing it with you. I was concerned of clearing my own db, but i can of course test with a db on another volume and leave this one intact.

I'll do this and once it's all working I'll let you know :v:

SamLubbers:
@aleesaan after some learning and digging, I have an update about this pull request :raised_hands: 

The way I had migrated the permissions for editing and deleting roles was actually correct! But... I made a silly mistake that I only just discovered: I removed the permissions for non-logged-in users to select roles. Because of that, you get that error (because the rolesSearch is a select statement on the roles database.

For now I will again export the whole database and replace the `init.sql` file. But soon I will upade this crappy setup with [proper migrations](https://hasura.io/docs/latest/graphql/core/migrations/index.html) which will make it much easier to share updates in the backend :ok_hand: 

If after my next commit you again remove the database volume (`docker volume rm docker_db_data`) then the frontend should work as expected, and you should have the correct configuration: new roles automatically get the correct id in the `authorId` column, and only logged in users can edit/delete their own roles (and not other roles)

aleesaan:
Sounds great!! I'm only wondering: can we create another user, so we can test this works as it should? With only one role, we can't test whether I can't edit/update someone else's roles.

SamLubbers:
@aleesaan good point :sweat_smile: I just made a new user with username `test` and password `test`. I'll soon write the authentication documentation so that you also know how to add users.

Now while I'm testing it I see that there is no good error handling: when you delete or edit a role of another user it doesn't show an error on the interface. I'll fix this now.

aleesaan:
Good that you spotted it! That's indeed why I thought to try with a different account :)  
  
As a side note, you might notice that the toast errors aren't being shown properly: I fixed that in my other branch with the rest of the "My Roles" changes! So don't worry about that, just make sure you handle the errors and dispatch the right actions.

src/GraphQL/clean.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * before sending data to the backend for an insert or update operation, we need
+ * to remove from the input those fields that cannot be updated. If we send those
+ * fields the backend will return an error as there is validation prohibiting the
+ * user to insert or update certain fields
+ */
+import { cloneDeep } from "lodash";
+import { Role } from "@/types/database";
+
+export function cleanRoleInput(roleInput: Role) {
```
aleesaan, line 13: We can achieve this quite easily using [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment):

```ts
export function cleanRoleInput(roleInput: Role) {
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  const { id, authorId, createdDate, ...role } = roleInput;

  return role;
}
```
Note that we need the `// eslint-disable...` cause it will complain of us not using the unneeded vars.  
  
Additionally, can you move this function to a `src/utils/cleanRoleInput.ts` file?

src/types/database.ts
```diff
@@ -0,0 +1,27 @@
+/*
+ * types and interfaces for data stored in the database
+ * */
+
+export interface Translation {
+  en: string;
+  nl: string;
+}
+
+export type Role = {
```
aleesaan, line 10: Instead of `types`, can we call the folder `models`? Also, can you create separate `role.ts` and `translation.ts` files? So we can keep things a bit more organized.

aleesaan (changes requested):

src/GraphQL/clean.ts
SamLubbers: that's a very neat one-liner :sparkles: 

I also moved the function to the `utils` directory. It indeed makes sense to have it there, so that in the GraphQL directory we only have graphql queries.

One question: why do you recommend making a file for only this function? The reason I made `clean.ts` in the first place is because I foresee us having many more functions that do a similar kind of operation. I thought we could then all group them in this file

aleesaan: I think this is quite specific logic, I'm not sure what else we could put in this file. I'd change the name for now, and we can change it back later, if see we can group more stuff.

aleesaan (approved):
Ok I think everything looks good!! I fixed a couple of minor things and the tests, so I can merge. I will then rebase my [My Roles PR](https://github.com/xrnl/aginun/pull/250), so we can test the full functionality, and fix anything we notice after that :)  
  
In general I couldn't really understand much of the changes you made in the docker files, do you think you could perhaps write a short summary of the (important/relevant) changes you made here?    

I'll merge it now, feel free to answer whenever :)

